PR #1, opened 2017-02-13
Add first draft of API

chriskuehl:


README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### If you only have the interpreter
+
+```python
+>>> identify.tags_from_interpreter('python3.5')
+{'python', 'python3', 'python3.5'}
+>>> identify.tags_from_interpreter('bash')
+{'shell', 'bash'}
+>>> identify.tags_from_interpreter('some-unrecognized-thing')
+{}
```
asottile: `set()`

README.md
```diff
@@ -0,0 +1,71 @@
...
+* Mode (is it executable?)
+* File name (mostly based on extension)
+* If executable, the shebang is read and the interpreter interpreted
+
+
+### If you only have the filename
+
+```python
+>>> identify.tags_from_filename('file.py')
+{'text', 'python'}
```
asottile: conventionally `.py` files are "text", but it's possible that it isn't.  Shouldn't the `text` tag only be applied to on-disk things?

asottile: ah I see, this is a performance optimization.  I think this is a good idea :shipit:

chriskuehl:
Perhaps instead of returning a set of tags it should return a namedtuple, in order to better represent that some of the tags are mutually exclusive, e.g.

```python
FileDetails(
    filename='blah.py',
    type=FILE,
    binary=False,
    executable=True,
    interpreter='/usr/bin/env python2.7',
    tags={'python2.7', 'python2', 'python'},
)
```

pre-commit would then have to be responsible for turning those into a flat set of tags like the README currently says in this branch.

PR #3, opened 2017-02-13
Add shebang parsing

chriskuehl:
This is almost an exact copy-paste from pre-commit:
https://github.com/pre-commit/pre-commit/blob/master/pre_commit/util.py

It kinda sucks that it's `identify.identify`. All of the code could fit in one module, but we're probably going to want the mappings from extension to tags and interpreter to tags to go in separate modules, so I made this a package.

Should I move this code to `identify/__init__.py` so you can still `import identify` instead of `from identify import identify`? That seems kinda sneaky.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,7 +9,6 @@ omit =
 
 [report]
 show_missing = True
-skip_covered = True
 
 exclude_lines =
     # Have to re-enable the standard pragma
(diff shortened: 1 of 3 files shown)


asottile:
I think there's going to be enough parts such that modules make sense.  I hate the `__init__.py` dance too heh.

PR #7, opened 2017-02-16
Implement tags_from_interpreter

chriskuehl:


tests/identify_test.py
```diff
@@ -85,6 +85,9 @@ def test_tags_from_filename(filename, expected):
 
 @pytest.mark.parametrize(('interpreter', 'expected'), (
     ('python', {'python'}),
+    ('python3', {'python3', 'python'}),
+    ('python3.5.2', {'python3', 'python'}),
+    ('/usr/bin/python3.5.2', {'python3', 'python'}),
```
asottile: can you add a case for `/usr/bin/herpderpderpderpderp`.  I have a sneaking suspicion it infinite loops

chriskuehl: added, though it doesn't infinite loop (it gets cut down to `herpderpderpderpderp` right at the start, then the call to `rpartition` makes it the empty string after it doesn't match on the first iteration)

PR #12, opened 2017-02-18
Add a bunch more extensions

chriskuehl:
These come from cloning the 50 most starred GitHub repos and picking out extensions that seem like they should be added.


identify/extensions.py
```diff
@@ -4,62 +4,138 @@
...
 
 EXTENSIONS = {
+    'apinotes': {'text', 'apinotes'},
+    'asar': {'binary', 'asar'},
+    'bash': {'text', 'shell', 'bash'},
     'bat': {'text', 'batch'},
     'bz2': {'binary', 'bzip2'},
     'c': {'text', 'c'},
     'cc': {'text', 'c++'},
+    'cfg': {'text', 'conf'},
```
asottile: This is probably going to change the parsing of setup.cfg

identify/extensions.py
```diff
@@ -4,62 +4,138 @@
...
+    'pdf': {'binary', 'pdf'},
+    'pem': {'text', 'pem'},
+    'php': {'text', 'php'},
+    'php4': {'text', 'php'},
+    'php5': {'text', 'php'},
+    'plist': {'text', 'plist'},
     'png': {'binary', 'image', 'png'},
     'po': {'text', 'pofile'},
     'pp': {'text', 'puppet'},
+    'proto': {'binary', 'proto'},
```
asottile: Hmmm i thought these were text files

chriskuehl: yup they are, oops

identify/extensions.py
chriskuehl: yeah, not sure what to do. obviously `.conf`/`.cfg`/`.cnf` represent tons of different file formats

maybe we should just omit it (or just have `{'text'}`) since you can't reliably assume anything based on the type?

asottile: I was meaning towards omission as well (why I didn't fill out  . txt).  Maybe we just leave these as text and give higher priority to the named files?

asottile: Leaning*

chriskuehl:
@asottile made filename matches take precedence over extension matches so that, for example, `.cfg` can be very generic, but `setup.cfg` can be more specific.

PR #20, opened 2017-07-14
Use rpartion instead of if + split

asottile:
I keep thinking about this but also keep forgetting to try it, let's see what travis says :)


chriskuehl:
lol nice indentation

asottile:
lol, github editor

PR #31, opened 2017-11-13
add file extension for saltstack state files

soostdijck:


asottile:
@soostdijck aren't `.sls` files actually jinja templates over yaml?

soostdijck:
That's true, they can have jinja(2) in there so I suppose their own filetype is a good idea.

asottile:
@soostdijck thanks again for the PR!

This has been included in v1.0.7

Issue #29, opened 2017-09-21
Aliasing system?


asottile:
I frequently forget whether it is `javascript` or `js` (or various others) -- perhaps we should have an aliasing system?

michaelfromyeg:
Is this ticket in progress? Perhaps it's worth re-visiting #155 as the simple solution, since (afaict from the issues) this doesn't seem to be a widespread problem.

If aliasing is a must-do, could I pick this up?

asottile:
I'm declining this

asottile:
6 years and clearly yagni. it doesn't pass silently so at most it's a cosmetic annoyance

PR #36, opened 2018-04-04
Add .wsgi extension

asottile:
For python wsgi applications it's relatively common to list your entrypoint in a `.wsgi` file

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -117,6 +117,7 @@ EXTENSIONS = {
     'whl': {'binary', 'wheel', 'zip'},
     'woff': {'binary', 'woff'},
     'woff2': {'binary', 'woff2'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -117,6 +117,7 @@
     'whl': {'binary', 'wheel', 'zip'},
     'woff': {'binary', 'woff'},
     'woff2': {'binary', 'woff2'},
+    'wsgi', {'text', 'wsgi', 'python'},
```
chriskuehl: derp, nice comma

asottile: lol gdi

asottile: that's what I get by contributing from github ui and not using `pre-commit` :'(

asottile:
I split out #29 a while back on designing some sort of alias system -- haven't really put much thought into it yet though!

PR #38, opened 2018-04-11
Added more new extension types

rob-orr:
Added Kotlin and XSL, and added `xml` type to XSD.


asottile:
v1.0.11 -- thanks again!

PR #42, opened 2018-05-05
Add perl to the list of interpreters

asottile:


asottile:
I used `find /usr/bin/ -type f -executable | xargs --replace bash -c 'head -1 {} | grep -q '"'"'^#!'"'"' && echo -n {}: && head -1 {}' | grep -Ev 'bin/(sh|bash|python)'`

PR #48, opened 2018-06-09
Ignore case for extensions

asottile:


asottile:
@chriskuehl is this a good idea?

tests/identify_test.py
```diff
@@ -85,6 +85,11 @@ def test_tags_from_path_binary(tmpdir):
         ('mod/test.py', {'text', 'python'}),
         ('mod/Dockerfile', {'text', 'dockerfile'}),
 
+        # case of extension should be ignored
+        ('f.JPG', {'binary', 'image', 'jpeg'}),
+        # but case of name checks should still be honored
+        ('dockerfile.py', {'text', 'python'}),
```
chriskuehl: nice

PR #50, opened 2018-07-09
Add Yang, Yin, MIB and XQuery extensions

iwanb:
Hi,

I would like to add a few extensions:
- Yang and Yin (XML representation of Yang) extensions, a modeling language for network equipment (https://tools.ietf.org/html/rfc7950#section-5.2)
- MIB extension, the modeling language for SNMP (mainly) (https://en.wikipedia.org/wiki/Management_information_base)
- XQuery extensions, an XML query language (http://docs.basex.org/wiki/XQuery_Extensions)

Iwan


asottile:
This is part of v1.1.2 🎉

iwanb:
Thank you!

PR #56, opened 2018-10-05
Add .cu extension for cuda

prem-nuro:
Intentionally excluded C because CUDA C has some additional syntax.


asottile:
Thanks!  This has been released as part of v1.1.7! :tada:

PR #60, opened 2018-11-09
Added the py, pydsl, and pyobjects interpreters for salt

raddessi:
Disabled the executable checks on files since not all python is required
to be executable


tests/identify_test.py
```diff
@@ -184,7 +219,7 @@ def test_parse_shebang_from_file_does_not_exist():
 def test_parse_shebang_from_file_nonexecutable(tmpdir):
     x = tmpdir.join('f')
     x.write_text('#!/usr/bin/env python', encoding='UTF-8')
-    assert identify.parse_shebang_from_file(x.strpath) == ()
+    assert identify.parse_shebang_from_file(x.strpath) == ('python',)
```
asottile: this change is inappropriate -- this test is *intentionally* demonstrating that files that are non-executable are never checked for shebangs

raddessi: So to respond to your comment since I removed that logic I have also removed the test for it

asottile: yes, and I'm telling you that's unacceptable

raddessi: Ah I'm sorry I only saw the inline comment and not the comment to the PR itself :) my fault

OK, why do you say this is unacceptable?

I thought I had left a comment about possibly adding some logic for only looking at the content for specific file types but I don't actually see that comment in here so my apologies, but yes for performances sake I would go that way myself.

raddessi: ah, I had left the comment in https://github.com/chriskuehl/identify/issues/61

raddessi: Is there a specific way you want to continue with this? Conditional inspection would also be nice for .mk files since those are not only used for Makefiles but python config files for Check-MK's version of nagios.

asottile: Maybe something similar to #63?

PR #62, opened 2019-01-02
Identify .j2 extension as jinja2

speshak:
Ansible uses .j2 as the de-facto extension for Jinja templates.  This change identifies .j2 files as jinja.


asottile:
thanks!

PR #69, opened 2019-02-14
Add Solidity file extension .sol

schmir:


asottile:
🎉 thanks! this has been released as 1.2.2

PR #76, opened 2019-04-27
add jsonnet file extension

sjansen:
Jsonnet is a JSON templating language;

 http://jsonnet.org


identify/extensions.py
```diff
@@ -68,6 +68,7 @@
     'jpg': {'binary', 'image', 'jpeg'},
     'js': {'text', 'javascript'},
     'json': {'text', 'json'},
+    'jsonnet': {'test', 'jsonnet'},
```
asottile: I think this is meant to be `text`?

sjansen: correct

asottile:
Thanks! This is part of 1.4.2

PR #77, opened 2019-05-10
Add both Dart and Zig file detection

haarts:
Trivial additions of course.


asottile:
this is part of v1.4.3 -- thanks again for the patch!

PR #78, opened 2019-06-11
Add .ics to list of extensions

marco-c:


asottile:
More specifically I think this should be icalendar?

marco-c:
> More specifically I think this should be icalendar?

It's not strictly specified what the tags should be (or I couldn't find where it is). Perhaps we can use both `calendar` and `icalendar`?

asottile:
> > More specifically I think this should be icalendar?
> 
> It's not strictly specified what the tags should be (or I couldn't find where it is). Perhaps we can use both `calendar` and `icalendar`?

`.ics` is specifically for icalender though, not a generic calendar file

marco-c:
> .ics is specifically for icalender though, not a generic calendar file

I thought tags specified some properties about files. E.g. `'jpg': {'binary', 'image', 'jpeg'}` says that jpg is a binary file, it's of type image, and in particular it's JPEG format.
Anyway, it's fine by me, so I've changed `calendar` to `icalendar`.

asottile:
released as 1.4.4 -- thanks for the contribution! :tada:

PR #82, opened 2019-07-20
Identify CMakeLists.txt as cmake file

geieredgar:
Currently, CMakeLists.txt files are identified as plain text files because of their .txt extension.
Instead they should be tagged with `cmake`.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -189,6 +189,7 @@ NAMES = {
     '.npmignore': {'text', 'npmignore'},
     '.yamllint': {'text', 'yaml', 'yamllint'},
     'AUTHORS': EXTENSIONS['txt'],
+    'CMakeLists.txt': EXTENSIONS['cmake'],
     'COPYING': EXTENSIONS['txt'],
     'Dockerfile': {'text', 'dockerfile'},
     'Gemfile': EXTENSIONS['rb'],


identify/extensions.py
```diff
@@ -203,3 +203,7 @@
     'Rakefile': EXTENSIONS['rb'],
     'setup.cfg': EXTENSIONS['ini'],
 }
+
+NAMES_IGNORE_EXTENSION = {
+    'CMakeLists.txt': EXTENSIONS['cmake'],
```
asottile: I believe the smaller change is to add `CMakeLists.txt` to `NAMES` -- is there a reason that doesn't work?

geieredgar: Thank you for the fast response!

It would work, but it would result in CMakeLists.txt files being tagged with 'plain-text' while a *.cmake file would not have such a tag.  
I believe CMakeLists.txt and *.cmake files should be treated the same, but I understand if you don't want to add the mapping for just this one special case. 

However, I would be fine with just adding `CMakeLists.txt` to `NAMES`. I will submit a new commit if you want.

asottile: yes, let's do that

asottile:
released as v1.4.6! thanks again! (sorry for the delay)

PR #83, opened 2019-08-12
Add nix file type

zimbatm:
Nix is a language and build system. See https://nixos.org/nix/


asottile:
released as v1.4.6! thanks again!

Issue #61, opened 2018-11-09
salt sls files with python shebangs are not properly recognized


raddessi:
Some salt sls files are written in python directly, I've added those interpreters in a PR here https://github.com/chriskuehl/identify/pull/60 as well as disabled the executable checks since these files are not required to be and also disabled the optimization for skipping peeking at the file contents if we recognize the extension.. I do see how this a nice speedup though.. maybe the extensions.EXTENSIONS dict should be changed from:
```
EXTENSIONS = {
...
    'sh': {'text', 'shell'},
    'sls': {'text', 'salt'},
    'so': {'binary'},
...
}
```
to something like:

```
EXTENSIONS = {
...
    'sh': {
        'tags': {'text', 'shell'},
        },
    'sls': {
        'tags': {'text', 'salt'},
        'always-peek': True,
    },
    'so': {
        'tags': {'binary'},
    },
...
}
```
By changing each value to its own dict which could have an optional key always-peek(?) so then you could .get('always-peek', False) to check if the file should be inspected?

raddessi:
Hmm, let me fix those failing unit tests first..

raddessi:
Should be happy now :)

PR #86, opened 2019-11-14
Added parsing for saltstack interpreted files

raddessi:
Hey! This is the follow up to https://github.com/chriskuehl/identify/issues/61 and the continuation of https://github.com/chriskuehl/identify/pull/60, I hope this is more inline with how you would like. Let me know what you think.

I'm not 100% positive of how different tags should be added, IE if all the salt fiels should be tagged with 'salt' and then additionally the 'salt-<renderer>' tag, for now I only added the non-default renderer tags since I think matching a default renderer tag when the contents of the file is not actually  readable by that renderer would cause issues. I'm on the fence though.


raddessi:
Also I'm not sure of a way to cleanly deal with the possible combination of any of these renderers as you are able to mix and match them at will, IE: `#! jinja|py`.. Should they just not be tagged?

raddessi:
Actually now that I think about it I do think this needs some work, but I'll need your input for direction. I've added some failing tests that illustrate the problem.. The test expectations should probably be set first though and I'm not sure about that. Thanks

asottile:
I think the approach is better than last time, I'm not sure the `#! foo|bar` syntax is something that `identify` should try and handle -- it's not at all standard and seems invented by saltstack (?)

let me know what you think, sorry for the brevity I'm trying to get through my backlog of email :S happy to elaborate on things!

raddessi:
Maybe it is better to keep tagging anything ending with .sls as `salt` only, but then add the extra interpreter tag when we match a single renderer? So `#! py` would get tagged with `salt` and `python` rather than trying to make a subset of mixed tags like `salt-py` etc?

asottile:
yeah I think that makes the most sense, and would be less of a breaking change from the current behaviour.  I think it's also fine / expected to not handle the `foo|bar` syntax since that's salt specific and imo not really in the spirit of this tool (to handle ever weird combination that other tools might do)

raddessi:
Hey there! I'm so sorry for letting this sit in your PR queue, I'll do my best to get this closed as quickly as possible. Check out this iteration, I think it is much cleaner. I wasn't able to make the code completely generic since the logic to parse shebang formats would probably change with different types but I think this is pretty decent.

One remaining issue is that coverage thinks some lines are not being tested and as far as I can see I think they are.. do you see an obvious issue? 

Thanks again for keeping this alive, I hope your year is going well :)

identify/identify.py
```diff
@@ -73,6 +75,42 @@ def tags_from_path(path):
...
+                        )
+                    ],
+                )
+            except (IndexError, KeyError):
+                pass
+        else:
+            # the default interpreter is jinja
+            ret.update(extensions.EXTENSIONS['jinja'])
+
+    return ret
```
raddessi, line 111: These are the lines coverage is complaining about missing tests. I'm stumped.

CAM-Gerlach (commented):
I'm just a random contributor, not @asottile , but I had some comments on your file type additions based on my experience.

With regard to your overall approach, it would simplify the implementation considerably, make it more flexible/general, and less brittle/special case, if you avoided implementing a bunch of bespoke logic just for parsing salt files (requiring any other extension added to `EXTENSIONS_NEED_SHEBANG_CHECK` to do the same, or silently not work properly). Instead, why not move the `salt` extension key/value as-is to the `EXTENSIONS_NEED_SHEBANG_CHECK` dict, add the appropriate interpreters to the interpreters file, add a function that returns whether the file needs a shebang check (e.g. via the first 3-4 lines of your existing new one), and tweak the execution flow to cut out the `else` on line 57 and change the `if` below to `if (not len(t) and executable) or needs_shebang_check(os.path.basename(path))`, so the shebang check runs on such files? This is also a little more efficient (the extra extension check only runs when the shebang isn't already known to be needed), matches the format of the existing `EXTENSIONS_NEED_BINARY_CHECK`, and avoids unnecessary bespoke code and data to retrieve extensions from tags and having to hardcode interpreters for each extension, and extensions for each interpreter. Plus, it completely avoids any of the codes the test coverage is complaining about.

identify/identify.py
raddessi: yep, I verified 4 locations get hit

identify/extensions.py
```diff
@@ -100,6 +101,7 @@
     'lr': {'text', 'lektor'},
     'lua': {'text', 'lua'},
     'm': {'text', 'c', 'objective-c'},
+    'mako': {'text', 'mako'},
```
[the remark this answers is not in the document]
raddessi: Right on, I threw a couple of these in only because I didn;t see them in there already and thought why not, but if there are issues I'll just pull mako no problem there.

identify/extensions.py
```diff
@@ -52,9 +52,10 @@
     'gif': {'binary', 'image', 'gif'},
     'go': {'text', 'go'},
     'gotmpl': {'text', 'gotmpl'},
+    'gpg': {'text', 'gnupg'},
```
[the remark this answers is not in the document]
raddessi: Great catch, I'll update when I get back

identify/extensions.py
```diff
@@ -179,6 +181,7 @@
     'tgz': {'binary', 'gzip'},
     'thrift': {'text', 'thrift'},
     'tiff': {'binary', 'image', 'tiff'},
+    'tmpl': {'text', 'cheetah'},
```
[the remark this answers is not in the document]
raddessi: ah I did tot find those on a quick search, good call again. I'll remove this one.

raddessi:
> I'm just a random contributor, not @asottile , but I had some comments on your file type additions based on my experience.
> 
> With regard to your overall approach, it would simplify the implementation considerably, make it more flexible/general, and less brittle/special case, if you avoided implementing a bunch of bespoke logic just for parsing salt files (requiring any other extension added to `EXTENSIONS_NEED_SHEBANG_CHECK` to do the same, or silently not work properly). Instead, why not move the `salt` extension key/value as-is to the `EXTENSIONS_NEED_SHEBANG_CHECK` dict, add the appropriate interpreters to the interpreters file, add a function that returns whether the file needs a shebang check (e.g. via the first 3-4 lines of your existing new one), and tweak the execution flow to cut out the `else` on line 57 and change the `if` below to `if (not len(t) and executable) or needs_shebang_check(os.path.basename(path))`, so the shebang check runs on such files? This is also a little more efficient (the extra extension check only runs when the shebang isn't already known to be needed), matches the format of the existing `EXTENSIONS_NEED_BINARY_CHECK`, and avoids unnecessary bespoke code and data to retrieve extensions from tags and having to hardcode interpreters for each extension, and extensions for each interpreter. Plus, it completely avoids needing any of the code the test coverage is complaining about.

There is a lot to unpack here :) Yeah I wasn't very happy about the setup here overall and I think this sounds like a good idea but I would have to think it over more. I'm currently on a work trip and will be knee deep in patch panels for the next week or so but as soon as I get back I'll look this over. Thanks for the reviews!

identify/identify.py
```diff
@@ -61,15 +62,27 @@ def tags_from_path(path: str) -> Set[str]:
...
     t = tags_from_filename(os.path.basename(path))
     if len(t) > 0:
         tags.update(t)
-    else:
-        if executable:
-            shebang = parse_shebang_from_file(path)
-            if len(shebang) > 0:
-                tags.update(tags_from_interpreter(shebang[0]))
+
+    ext = os.path.splitext(os.path.split(path)[-1])[-1].lstrip('.')
```
raddessi, line 70: I don't love that this introduces a need to parse the extension on all files.. but I don't see a way around that when the logic to determine the extra requirement is by extension. Thoughts?

identify/identify.py
```diff
@@ -61,15 +62,27 @@ def tags_from_path(path: str) -> Set[str]:
...
+        except KeyError:
+            pass
+
+        shebang = parse_shebang_from_file(path)
+        if len(shebang) > 0:
+            tags.update(
+                tags_from_interpreter(
+                    shebang[0].split('|')[0].strip(),
+                ),
+            )
```
raddessi, line 85: The splitting of the `|` delimited interpreters may be better located in the `_shebang_split` function.. I can move there if this looks good.

identify/identify.py
```diff
@@ -203,7 +216,14 @@ def parse_shebang_from_file(path: str) -> Tuple[str, ...]:
...
         raise ValueError(f'{path} does not exist.')
-    if not os.access(path, os.X_OK):
+    ext = os.path.splitext(os.path.split(path)[-1])[-1].lstrip('.')
+    if (
+        ext not in extensions.EXTENSIONS_NEED_INTERPRETER_CHECK and
+        not os.access(
+            path,
+            os.X_OK,
+        )
+    ):
```
raddessi, line 226: This is the other place the extension is needed.. should it be passed to the function instead of re-parsed?

identify/identify.py
```diff
@@ -257,7 +277,7 @@ def license_id(filename: str) -> Optional[str]:
             return spdx
 
         # skip the slow calculation if the lengths are very different
-        if norm and abs(len(norm) - len(norm_license)) / len(norm) > .05:
+        if norm and abs(len(norm) - len(norm_license)) / len(norm) > 0.05:
             continue
```
raddessi, line 281: I didn't make a few of these changes, they were done via precommit or black.. Maybe the precommit config changed recently?

tests/identify_test.py
```diff
@@ -145,6 +145,47 @@ def test_tags_from_path_plist_text(tmpdir):
...
+        ('genshi', {'salt', 'file', 'non-executable', 'text'}),
+        ('gpg', {'salt', 'file', 'non-executable', 'text', 'gnupg'}),
+        ('jinja', {'salt', 'file', 'non-executable', 'text', 'jinja'}),
+        ('jinja|py', {'salt', 'file', 'non-executable', 'text', 'jinja'}),
+        ('jinja|yaml', {'salt', 'file', 'non-executable', 'text', 'jinja'}),
+        (
+            'jinja|yaml|gpg', {
+                'salt', 'file', 'non-executable', 'text', 'jinja',
+            },
+        ),
```
raddessi, line 161: This does have the benefit of files with the extension `.gpg` being marked as binary while files with the `gpg` interpreter still get marked as text.

raddessi (commented):
@CAM-Gerlach Thank you again for your review :) I started working on your recommendations and saw another way to clean it up a bit more so I did that to get both of your thoughts on it.

It may be over reaching to add these interpreters since they aren't standards but I think it at least makes more sense than the previous iterations. I left a few comments on points I'd like opinions on if you have time. Thanks again

tests/identify_test.py
raddessi: I apologize, I quotes the wrong section. The relevant line is 153:

`('gpg', {'salt', 'file', 'non-executable', 'text', 'gnupg'}),`

identify/identify.py
CAM-Gerlach: If the `|` convention to separate multiple interpreters is non-standard and only used with one particular file type, I don't know about trying to parse it on all files. I can't find much Googling about other uses, but the [one](http://reasonableapproximation.net/2010/07/27/silly-things-to-do-with-shebang-lines.html) usage I can find has different syntax than salt uses. But that's ultimately up to @asottile .

identify/identify.py
CAM-Gerlach: I would suggest eliminating this check entirely unless there's something I'm missing (which is entirely possible); see below.

identify/identify.py
```diff
@@ -203,7 +216,14 @@ def parse_shebang_from_file(path: str) -> Tuple[str, ...]:
...
-    if not os.access(path, os.X_OK):
+    ext = os.path.splitext(os.path.split(path)[-1])[-1].lstrip('.')
+    if (
+        ext not in extensions.EXTENSIONS_NEED_INTERPRETER_CHECK and
+        not os.access(
+            path,
+            os.X_OK,
+        )
+    ):
         return ()
```
CAM-Gerlach, line 228:
```suggestion
```

@asottile The only function that calls this already checks this same condition before running `parse_shebang_from_file()`, so is there a reason it needs to be checked twice? There might be something I'm missing, but if not, eliminating the redundant check would reduce complexity and increase performance without an obvious downside, and it also prevents any other third-party consumers from using it on files not marked executable if that restriction doesn't make sense for their application.

tests/identify_test.py
CAM-Gerlach: Nice!

CAM-Gerlach (commented):
Thanks! I had a few comments on your implementation; the rest is up to @asottile .

raddessi:
FYI only, I'm playing around with this still. I don't really love any of these approaches so far as there is either a spaghetti code path or some duplicated processing.. I'm playing with either a custom path object with `@property` getters to cache the duplicated calls or just a `functools.lru_cache` on top of some functions.. I need to do some performance comparisons before I send anything up. I don't want to add duplicated calls but I also don't want to add a ton of changes just to avoid them either.. I'm still on the fence about the whole path forward.

kotki4017 (approved):

PR #87, opened 2019-12-04
Handle malformed shebangs

ngie-eign:
One of the tests in valgrind specifically provides a shebang that is
malformed (contains just a `#!`), which is technically valid bourne
shell syntax.

Fix the code to check that the result from `shlex.split(..)` isn't an
empty `tuple` to avoid an `IndexError` traceback when malformed input
is provided to `identify`.

Add a testcase for the issue.

### DIFF
--- a/tests/identify_test.py
+++ b/tests/identify_test.py
@@ -54,6 +54,15 @@ def test_tags_from_path_simple_file(tmpdir):
     }
 
 
+def test_tags_from_path_file_with_incomplete_shebang(tmpdir):
+    x = tmpdir.join('test')
+    x.write_text('#!   \n', encoding='UTF-8')
+    make_executable(x.strpath)
+    assert identify.tags_from_path(x.strpath) == {
+        'file', 'text', 'executable',
+    }
+
+
 def test_tags_from_path_file_with_shebang_non_executable(tmpdir):
     x = tmpdir.join('test')
     x.write_text('#!/usr/bin/env python\nimport sys\n', encoding='UTF-8')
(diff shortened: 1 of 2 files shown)


tests/identify_test.py
```diff
@@ -54,6 +54,14 @@ def test_tags_from_path_simple_file(tmpdir):
     }
 
 
+def test_tags_from_path_file_with_incomplete_shebang(tmpdir):
+    x = tmpdir.join('test')
+    x.write_text('#!   \n', encoding='UTF-8')
+    assert identify.tags_from_path(x.strpath) == {
+        'file', 'text', 'non-executable',
+    }
```
asottile: I don't think this test exercises this change -- the shebang code isn't run unless the file is made executable

ngie-eign: Ah.... ok :D.

ngie-eign: Fixed -- thank you for that note :D!

ngie-eign:
Hah... Forgot to change my commit email address. I really need to come up with a better way to handle that in an automated fashion. Fixed :D.

asottile:
thanks for the fix!  this has been released as 1.4.8

ngie-eign:
@asottile: I need to figure out where you find your dog coding gifs :D...

asottile:
they come from https://github.com/chriskuehl/shipit ! (there's a user script I wrote in there that automatically posts them when clicking [approve]) -- most of them are cats, RNG just blessed you with doggos :D

PR #88, opened 2019-12-23
Add Bazel, Scheme and Idris filetypes

verhovsky:


identify/extensions.py
```diff
@@ -190,6 +194,8 @@
     '.npmignore': {'text', 'npmignore'},
     '.yamllint': {'text', 'yaml', 'yamllint'},
     'AUTHORS': EXTENSIONS['txt'],
+    'BUILD.bazel': {'text', 'bazel'},
+    'BUILD': {'text', 'bazel'},
```
asottile: other two extensions look good -- `BUILD` might be not-specific-enough -- how common is BUILD vs something with a `.bazel` extension?  could the `.bazel` extension be added above instead of here maybe?

verhovsky: I've only ever seen `BUILD` files. I just saw `BUILD.bazel` in https://github.com/FelixSeptem/pre-commit-golang/blob/master/.pre-commit-hooks.yaml 

There's also `WORKSPACE` files and `*.bzl` files.

And you could mark Bazel files as Python files, but it's techincally not Python, see https://docs.bazel.build/versions/master/skylark/language.html

asottile: ok, let's go with this and if someone complains we can always back it out later 👍

asottile:
oops forgot to click approve, here have a gif anyway :D

PR #91, opened 2020-01-25
Test with Python 3.7, 3.8, and PyPy3

scop:


asottile:
thanks, but until there's code changes this doesn't actually provide additional coverage (and just makes CI slower), when I switch this to AP I'll change what versions it targets

scop:
I disagree. It adds test coverage for 3 entire versions of python that are not covered by CI (or local tox runs by default) at the moment.

Also, for example there certainly are people using this with 3.8, and at the moment there's zero indication if this would work with it or not. setup.cfg doesn't mention, and tests are not being run. Ditto pypy3, although much more rare obviously. And if this works with 3.7, why not run public tests to guard/show it?

Is CI slowness an actual consideration? It's a matter of couple of minutes on Travis, probably much much, much less locally. `skip_missing_interpreters = True` would be a good idea for local runs, though.

PR #92, opened 2020-01-25
Recognize .pm as perl

scop:


asottile:
this has been released as part of v1.4.11 -- thanks for the patch!

PR #93, opened 2020-03-01
WIP: Emacs tags support

scop:
Here go first steps towards Emacs tags support. These are strings like `-*- shell-script -*-` that are fairly commonly found in files here and there.

Not quite complete yet, but filing the PR already now for an ack that this would be a desirable addition and comments.

This looks at the couple of first lines of files only, even though IIRC the tags can be located pretty much anywhere in a file. It's common for them to be at beginning of files though.

### DIFF
--- a/identify/emacs_tags.py
+++ b/identify/emacs_tags.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+EMACS_TAGS = {
+    'sh': {'shell', 'sh'},
+    'shell-script': {'shell', 'sh'},
+    'cperl': {'perl'},
+    'perl': {'perl'},
+    'python': {'python'},
+    # TODO a lot mode
+}
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -10,6 +10,7 @@ import shlex
 import string
 import sys
 
+from identify import emacs_tags
 from identify import extensions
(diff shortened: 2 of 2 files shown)


asottile:
I believe this would require reading all of every file always which isn't going to be acceptable

scop:
I believe looking at the couple of first lines would suffice for this to be very useful already.

asottile:
the design of identify is to be as fast as possible, this means avoiding I/O whenever possible

identify does not intend to identify all possible weird combinations that could be set up and I believe emacs tags fall into that category

do you have a real example of a file which currently isn't being identified? -- in all likelihood we should be adding more conventions / extensions and not slow parsing

scop:
Pretty much the whole completions/ subdir of https://github.com/scop/bash-completion/ serves as an example. I've worked around it in the repo's pre-commit config, but I can see non-detection happening for example for all shell libraries that are meant to be sourced that lack a (redundant) `*.sh` or the like filename extension or a (redundant and "wrong") executable bit+shebang.

scop:
Emacs tags isn't the only thing, there are similar `# ex` and `# vi` constructs for vi and friends etc etc. I understand why not everyone would be thrilled by a change like this, but "fixing" it by renaming e.g. to *.sh comes with a cost of its own, too.

asottile:
at least for pre-commit, the ~usual solution for this situation is `types: [text]` + `files: ^path/to/files/`

PR #94, opened 2020-03-10
Recognize jsonschema as json

nijel:


asottile:
do you have a reference for `.jsonschema` being an extension? everywhere I've seen uses `.json` for jsonschema files

nijel:
Not really. It was submitted by a contributor this way to my project and I thought it's a standard way. Looking into this topic now, there doesn't seem to be any standard in this area defined (for example https://stackoverflow.com/q/9391370/225718), so sticking with just `.json` is probably better.

asottile:
closing per above -- thanks for the patch nonetheless!

PR #102, opened 2020-04-11
support whatever.BUILD bazel files

TheButlah:
It's common for external bazel workspaces to use `external-project.BUILD` files, as opposed to `BUILD.bazel` files typically present in bazel projects. This is to differentiate them from the local workspace's bazel files, which have the format `BUILD` or `BUILD.bazel`. Hence, it would be good to treat these files like bazel files as well.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -206,6 +206,7 @@ NAMES = {
     '.yamllint': EXTENSIONS['yaml'] | {'yamllint'},
     '.zshrc': EXTENSIONS['zsh'],
     'AUTHORS': EXTENSIONS['txt'],
+    '.BUILD': {'text', 'bazel'},
     'BUILD.bazel': {'text', 'bazel'},
     'BUILD': {'text', 'bazel'},
     'CMakeLists.txt': EXTENSIONS['cmake'],


identify/extensions.py
```diff
@@ -206,6 +206,7 @@
     '.yamllint': EXTENSIONS['yaml'] | {'yamllint'},
     '.zshrc': EXTENSIONS['zsh'],
     'AUTHORS': EXTENSIONS['txt'],
+    '.BUILD': {'text', 'bazel'},
```
asottile, line 209: this mapping is for conventional names, the extensions mapping is above -- I don't think this patch does what you want 🤔

asottile: or, maybe I'm missing the intention here -- maybe you could add a test or two to demonstrate this :)

TheButlah: To be honest, I don't understand at all how this repo works - this is merely pattern matching on my part. My intent is to detect files ending in `.BUILD` and classify them the same way as `BUILD.bazel` or `BUILD` files.

TheButlah: What would be the best way to get such a change added?

asottile: I would suggest writing a test demonstrating what you want to do, and then adjusting this value until the test passes -- I'm not sure this is the correct place or value for what you're trying to do and a test should help demonstrate that.

I think the first mapping is more appropriate, but I also think it does case-insensitive matching on extensions so it's maybe not quite a perfect fit.  It's possible that what you have written is correct, but it would be better from a pull request perspective to demonstrate that precisely

TheButlah:
Closing for now. Might reopen if I put in the time for a proper PR rather than a quick file edit.

PR #105, opened 2020-05-11
Run pre-commit autoupdate

asottile:
Committed via https://github.com/asottile/all-repos


asottile:
hmmm, python2 and python3.5 🤔

PR #108, opened 2020-05-28
Add csh to interpreters and .cshrc to extensions

abduhbm:


identify/interpreters.py
```diff
@@ -15,4 +15,5 @@
     'sh': {'shell', 'sh'},
     'tcsh': {'shell', 'tcsh'},
     'zsh': {'shell', 'zsh'},
+    'csh': {'shell', 'csh'},
```
asottile, line 18: could you sort these, please? thanks!

abduhbm: Sorted. Thanks,

PR #117, opened 2020-07-18
feat: Detect common haskell suffixes (hs,lhs,chs)

merc1031:


identify/extensions.py
```diff
@@ -86,6 +88,7 @@
     'kml': {'text', 'kml', 'xml'},
     'kt': {'text', 'kotlin'},
     'less': {'text', 'less'},
+    'lhs': {'text', 'haskell'},
```
asottile, line 91: is it common that linters / formatters would handle all three extensions the same? or should these produce a different tag

merc1031: It might be nice to have them separate. i just didn't know if that would be ... polluting the namespace too much.
They are quite a bit different.
Literate haskell lets you mix in compilable code with like a blog almost
c haskell does crazy C processing inline with the haskell itself.

I am happy to split them up!

asottile: probably makes the most sense to split them if I had to guess

identify/extensions.py
```diff
@@ -16,6 +16,7 @@
     'cc': {'text', 'c++'},
     'cu': {'text', 'cuda'},
     'cfg': {'text'},
+    'chs': {'text', 'c2hs'},
```
merc1031, line 19: Im not totally happy with the lack of symmetry with the other 2 haskell variants below, but any variant of 'c2-haskell', 'c-2-haskell', 'c-to-haskell' just looked off to me

PR #118, opened 2020-07-20
Add common HDL language extensions

mattofak:
Add support for detecting files of the common hardware description
languages of verilog and VHDL.


asottile:
can you link a reference, please?

mattofak:
Sure; I altered the commit to point to the wikipedia articles for these sources files; and since wikipedia and Xilinx disagree for VHDL, I linked to a user guide for Vivado (Xilinx's IDE.)

I also noted that I forgot to include the common extensions for the header files, so those were added as well.

PR #127, opened 2020-09-04
add: nix-shell interpreter

blaggacao:
see: https://github.com/cachix/pre-commit-hooks.nix/issues/35

### DIFF
--- a/identify/interpreters.py
+++ b/identify/interpreters.py
@@ -6,6 +6,7 @@ INTERPRETERS = {
     'bash': {'shell', 'bash'},
     'csh': {'shell', 'csh'},
     'dash': {'shell', 'dash'},
+    'nix-shell': {'shell', 'bash'},
     'node': {'javascript'},
     'nodejs': {'javascript'},
     'perl': {'perl'},


identify/interpreters.py
```diff
@@ -6,6 +6,7 @@
     'bash': {'shell', 'bash'},
     'csh': {'shell', 'csh'},
     'dash': {'shell', 'dash'},
+    'nix-shell': {'shell', 'bash'},
```
roberth, line 9: This is suitable as a first-order approximation, because most of the time it will be `-i bash` indeed.
It'd be better to write a function to parse the consecutive `#!nix-shell` lines to find the `-i` value. You could call the new function from around here https://github.com/chriskuehl/identify/blob/62e98d7bee975f686bab8257c4fe95ff0ff7ab80/identify/identify.py#L162

blaggacao: Thanks! In other words: please.merge and follow up with another PR. :wink:

asottile, line 9: iirc (it was proposed before) this doesn't actually work since `nix-shell` could be any interpreter

blaggacao: I think you are referring to the cases of https://github.com/chriskuehl/identify/issues/128 where `-i <interpreter>`? Afaik, `nix-shell` does always invoke a bash environment. Wold this addition conflict with #128? @roberth, you are more of an expert on the subject, could you clarify?

asottile: nah this was a long time ago, nix does a strange thing where the actual language is on the second line of the file and the file is not always executable.  I can't find the issue, it appears it wasn't on this project but perhaps on one of the nix projects

blaggacao: https://gist.github.com/travisbhartwell/f972aab227306edfcfea

blaggacao:
Reopened, since this is ok and plays well with #130 for the wired things happening on the second line.

blaggacao:
So this is wrong, see: https://github.com/blaggacao/identify/blob/patch-4/tests/identify_test.py

PR #130, opened 2020-09-04
imp: add nix-shell interpreter parser

### LINKED ISSUE #128: Exotic shebang support
`#!.*nix-shell.*-i bash.*` is a valid `nix-shell` second-line shebang which should be parsed to `bash` interpreter (similarly `python`, `ruby`, etc.). 

@roberth maybe you can educate those who didn't know this form what is it all about?

See also https://github.com/chriskuehl/identify/pull/127#discussion_r483459849

and also: https://github.com/chriskuehl/identify/pull/127#discussion_r483758495

blaggacao:
nix-shell uses the second line to pass relevant arguments
to the interpreter. Linux (unlike BSD) 
-- see: https://unix.stackexchange.com/a/63981 --
only passes a single argument to the shebang command
(in this case env or nix-shell).

On the second line, if specified, the interpreter is passed
to nix-shell via the -i flag, so we parse this interpreter
if the second line is specified. If not (nix-shell called w/o
arguments) the implicit interpreter is bash.
-- see also: #127 --

closes #128 

@roberth & @zimbatm might I ask you for a quick review to ensure this implementation is 100% correct? Note: we might need to make sure `nix-shell` cmd is preserved (as a wrapper cmd for this use case) on `nix` 2.0

---

See also the nix manual https://nixos.org/manual/nix/stable/#use-as-a-interpreter

---

There might be additional information in this discussion: https://discourse.nixos.org/t/identify-nix-shell-shebangs-for-pre-commit/8888

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -159,6 +159,33 @@ def parse_shebang(bytesio):
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
         cmd = cmd[1:]
+        if cmd[:1] == ('nix-shell',):
+            if cmd[1:]:
+                return ()
+            multiple_shebangs = False
+            while bytesio.read(2) == b'#!':
+                multiple_shebangs = True
+                next_line = bytesio.readline()
+                try:
+                    next_line = next_line.decode('UTF-8')
+                except UnicodeDecodeError:
+                    return ()
+
+                for c in next_line:
+                    if c not in printable:
+                        return ()
+
+                line_tokens = tuple(shlex.split(next_line.strip()))
+                for i, token in enumerate(line_tokens):
+                    if not token == '-i':
+                        continue
+                    try:
+                        # the argument to -i flag
+                        cmd = (line_tokens[i + 1],)
+                    except IndexError:
+                        return ()
+            if not multiple_shebangs:
+                return ()
     return cmd
 
 
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -141,6 +141,27 @@ def _shebang_split(line):
         return line.split()
 
 
+def _parse_nix_shebang(bytesio, cmd):
+    while bytesio.read(2) == b'#!':
+        next_line = bytesio.readline()
+        try:
+            next_line = next_line.decode('UTF-8')
+        except UnicodeDecodeError:
+            return cmd
+
+        for c in next_line:
+            if c not in printable:
+                return cmd
+
+        line_tokens = tuple(_shebang_split(next_line.strip()))
+        for i, token in enumerate(line_tokens[:-1]):
+            if token != '-i':
+                continue
+            # the argument to -i flag
+            cmd = (line_tokens[i + 1],)
+    return cmd
+
+
 def parse_shebang(bytesio):
     """Parse the shebang from a file opened for reading binary."""
     if bytesio.read(2) != b'#!':
@@ -159,33 +180,8 @@ def parse_shebang(bytesio):
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
         cmd = cmd[1:]
-        if cmd[:1] == ('nix-shell',):
-            if cmd[1:]:
-                return ()
-            multiple_shebangs = False
-            while bytesio.read(2) == b'#!':
-                multiple_shebangs = True
-                next_line = bytesio.readline()
-                try:
-                    next_line = next_line.decode('UTF-8')
-                except UnicodeDecodeError:
-                    return ()
-
-                for c in next_line:
-                    if c not in printable:
-                        return ()
-
-                line_tokens = tuple(shlex.split(next_line.strip()))
-                for i, token in enumerate(line_tokens):
-                    if not token == '-i':
-                        continue
-                    try:
-                        # the argument to -i flag
-                        cmd = (line_tokens[i + 1],)
-                    except IndexError:
-                        return ()
-            if not multiple_shebangs:
-                return ()
+        if cmd == ('nix-shell',):
+            return _parse_nix_shebang(bytesio, cmd)
     return cmd
 
 
--- a/tests/identify_test.py
+++ b/tests/identify_test.py
@@ -219,57 +219,61 @@ def test_file_is_text_does_not_exist(tmpdir):
         (b"#! /path'with/quotes y ", ("/path'with/quotes", 'y')),
         # Test nix-shell specialites with shebang on second line
         (
-            b'#! /usr/bin/env nix-shell\n' +
+            b'#! /usr/bin/env nix-shell\n'
             b'#! nix-shell -i bash -p python',
             ('bash',),
         ),
         (
-            b'#! /usr/bin/env nix-shell\n' +
+            b'#! /usr/bin/env nix-shell\n'
             b'#! nix-shell -i python -p coreutils',
             ('python',),
         ),
         (
-            b'#! /usr/bin/env nix-shell\n' +
+            b'#! /usr/bin/env nix-shell\n'
             b'#! nix-shell -p coreutils -i python',
             ('python',),
         ),
         # multi-line and no whitespace variation
         (
-            b'#! /usr/bin/env nix-shell\n' +
-            b'#! nix-shell -p coreutils\n' +
+            b'#! /usr/bin/env nix-shell\n'
+            b'#! nix-shell -p coreutils\n'
             b'#! nix-shell -i python',
             ('python',),
         ),
         (
-            b'#! /usr/bin/env nix-shell\n' +
-            b'#!nix-shell -p coreutils\n' +
+            b'#! /usr/bin/env nix-shell\n'
+            b'#!nix-shell -p coreutils\n'
             b'#!nix-shell -i python',
             ('python',),
         ),
         (
-            b'#! /usr/bin/env nix-shell\n' +
+            b'#! /usr/bin/env nix-shell\n'
             b'#!\xf9\x93\x01\x42\xcd',
-            (),
+            ('nix-shell',),
         ),
         (
-            b'#! /usr/bin/env nix-shell\n' +
+            b'#! /usr/bin/env nix-shell\n'
             b'#!\x00\x00\x00\x00',
-            (),
+            ('nix-shell',),
         ),
         # non-proper nix-shell
-        # (b'#! /usr/bin/nix-shell', ()),  # out of scope test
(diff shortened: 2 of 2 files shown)


tests/identify_test.py
```diff
@@ -217,6 +217,9 @@ def test_file_is_text_does_not_exist(tmpdir):
         (b"#!/path'with/quotes    y", ("/path'with/quotes", 'y')),
         # Don't regress on leading/trailing ws
         (b"#! /path'with/quotes y ", ("/path'with/quotes", 'y')),
+        # Test nix-shell specialites with shebang on second line
+        (b"#! /usr/bin/env nix-shell\n#! nix-shell -i bash -p python", ('bash',)),
+        (b"#! /usr/bin/env nix-shell\n#! nix-shell -i python -p coreutils", ('python',)),
```
asottile, line 222: what about `-p python -i bash`

asottile (commented):
pretty sure this also won't work because nix-shell files aren't conventionally executable

blaggacao:
> pretty sure this also won't work because nix-shell files aren't conventionally executable

Do you remember the source for this claim? &mdash; it would surprise me, since what's the shebang for, then.

identify/identify.py
```diff
@@ -159,6 +159,15 @@ def parse_shebang(bytesio):
...
         cmd = cmd[1:]
+    if cmd and cmd[0].endswith('nix-shell'):
+        if bytesio.read(2) != b'#!':
+            return cmd
+        second_line = bytesio.readline().decode('UTF-8')
+        cmd = tuple(_shebang_split(second_line.strip()))
+        for i, token in enumerate(cmd):
+            if not token == '-i':
+                continue
+            cmd = (cmd[i + 1],)  # the argument to -i flag
```
roberth, line 170: Indeed it will take the last one.

tests/identify_test.py
```diff
@@ -217,6 +217,13 @@ def test_file_is_text_does_not_exist(tmpdir):
         (b"#!/path'with/quotes    y", ("/path'with/quotes", 'y')),
         # Don't regress on leading/trailing ws
         (b"#! /path'with/quotes y ", ("/path'with/quotes", 'y')),
+        # Test nix-shell specialites with shebang on second line
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -i bash -p python', ('bash',)),
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -i python -p coreutils', ('python',)),
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -p coreutils -i python', ('python',)),
+        (b'#! /usr/bin/env nix-shell', ('nix-shell',)),
+        # edge case - normally nix-shell is on an opaque nix path and accessed through env
+        (b'#! /usr/bin/nix-shell\n#! nix-shell -p coreutils -i python', ('python',)),
```
roberth, line 226: Some more test cases for multi-line, no space after `#!` and quotes.

```suggestion
        (b'#! /usr/bin/env nix-shell\n#! nix-shell -p coreutils -i python', ('python',)),
        # multi-line and no whitespace variation
        (b'#! /usr/bin/env nix-shell\n#! nix-shell -p coreutils\n#! nix-shell -i python', ('python',)),
        (b'#! /usr/bin/env nix-shell\n#!nix-shell -p coreutils\n#!nix-shell -i python', ('python',)),
        # interpret quotes correctly
        (b'#!/usr/bin/env nix-shell\n#!nix-shell --argstr x "a -i python3 p"\n#!nix-shell -p hello\n#!nix-shell -i bash\n#!nix-shell --argstr y "b -i runhaskell q"', ('bash',)),
```

nix-shell supports quotes. Here's [Nix's implementation](https://github.com/NixOS/nix/blob/8a945d6ddb0676b454458e6fe0e9ea6f8b4b5659/src/nix-build/nix-build.cc#L26) which claims to be inspired by Perl's `shellwords`.
Although python's [`shellwords`](https://github.com/mozillazg/python-shellwords/blob/master/shellwords/__init__.py#L43) package seems like a match, with `parse_env` on and `parse_backtick` off. Nix seems to allow env interpolation somewhere after `shellwords`.

identify/identify.py
blaggacao: What do you mean? Sorry, I couldn't make sense of this comment :confused: .

tests/identify_test.py
```diff
@@ -217,6 +217,13 @@ def test_file_is_text_does_not_exist(tmpdir):
         (b"#!/path'with/quotes    y", ("/path'with/quotes", 'y')),
         # Don't regress on leading/trailing ws
         (b"#! /path'with/quotes y ", ("/path'with/quotes", 'y')),
+        # Test nix-shell specialites with shebang on second line
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -i bash -p python', ('bash',)),
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -i python -p coreutils', ('python',)),
+        (b'#! /usr/bin/env nix-shell\n#! nix-shell -p coreutils -i python', ('python',)),
+        (b'#! /usr/bin/env nix-shell', ('nix-shell',)),
```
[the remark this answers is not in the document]
blaggacao: @asottile Do we have an api for notifying broken scripts? Else, I'd just `return ()`, ok?

identify/identify.py
roberth: The last occurrence of `-i`, so that's good.

tests/identify_test.py
blaggacao: Ok, I forgot about the multi-line case, will implement it.

tests/identify_test.py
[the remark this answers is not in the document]
asottile: yeah if we can't identify from the interpreter, just return `()`

identify/identify.py
```diff
@@ -159,6 +159,15 @@ def parse_shebang(bytesio):
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
         cmd = cmd[1:]
+    if cmd and cmd[0].endswith('nix-shell'):
```
asottile, line 162: should this be `if cmd == ('nix-shell',):`

identify/identify.py
```diff
@@ -159,6 +159,15 @@ def parse_shebang(bytesio):
...
+        if bytesio.read(2) != b'#!':
+            return cmd
+        second_line = bytesio.readline().decode('UTF-8')
```
asottile, line 165: will need an additional check for non-UTF-8 and non-readable as above

identify/identify.py
```diff
@@ -159,6 +159,15 @@ def parse_shebang(bytesio):
...
+        cmd = tuple(_shebang_split(second_line.strip()))
```
asottile, line 166: I believe we'll actually want to use `shlex.split` here since nix supports proper quoting

identify/identify.py
asottile, line 170: this could go out-of-bounds, can you add a test where `-i` is last

identify/identify.py
blaggacao: Thanks for this catch! Solved in https://github.com/chriskuehl/identify/pull/130/commits/00f56a68ea5d6ace7ee362ea5f3bd918ce2b204a

identify/identify.py
blaggacao: fixed in https://github.com/blaggacao/identify/commit/1595423afadb8c3f137b2371c31a9abcf7f1d6c7

identify/identify.py
blaggacao: solved in https://github.com/chriskuehl/identify/pull/130/commits/2a45f63977fb30fd2abd569f4488cd47549ce39a

identify/identify.py
```diff
@@ -159,7 +159,9 @@ def parse_shebang(bytesio):
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
         cmd = cmd[1:]
-        if cmd and cmd == 'nix-shell':
+        if cmd == ('nix-shell',):
+            if cmd[1:]:
+                return ()
```
blaggacao: This is bullshit...

identify/identify.py
```diff
@@ -162,26 +162,29 @@ def parse_shebang(bytesio):
...
-            for c in second_line:
-                if c not in printable:
-                    return ()
-
-            cmd = tuple(_shebang_split(second_line.strip()))
-            for i, token in enumerate(cmd):
-                if not token == '-i':
-                    continue
+            multiple_shebangs = false
+            while bytesio.read(2) != b'#!':
```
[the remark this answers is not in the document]
blaggacao: solved in https://github.com/chriskuehl/identify/pull/130/commits/be23107269babb2a604437fca7f9a80ba8094411 &mdash; please validate and resolve.

identify/identify.py
blaggacao: solved in https://github.com/chriskuehl/identify/pull/130/commits/be23107269babb2a604437fca7f9a80ba8094411 &mdash; please validate and resolve.

blaggacao:
> I get an email every time you push and it's failing the basic tests / linting

Oh, sorry. That side effect didn't occur to me. I also didn't think it would get that involve... Now it's green locally.

blaggacao:
@roberth Can you find another test to break this? :wink:

asottile:
after reading this a little more closely, it makes more sense in the error cases to return whatever first-line shebang that's been determined so far

I've adjusted the tests and code as well as split out the nix stuff to a helper function

blaggacao:
> it makes more sense in the error cases to return whatever first-line shebang that's been determined so far

Cool, yes I understand. It should not so much be of `identify`'s concern to detect errors.
Being a little more naive, the code looks also way cleaner.

LGTM

identify/identify.py
```diff
@@ -141,6 +141,27 @@ def _shebang_split(line):
...
+        try:
+            next_line = next_line.decode('UTF-8')
+        except UnicodeDecodeError:
+            return cmd
+
+        for c in next_line:
+            if c not in printable:
+                return cmd
+
+        line_tokens = tuple(_shebang_split(next_line.strip()))
```
blaggacao, line 156: Why did you revert to a potential fallback on line split?

blaggacao (commented):
Maybe just an oversight?

identify/identify.py
asottile: I misremembered how this used to work, and shlex.split can raise

I originally though this was a more true implementation of shebang parsing which would be `line.split(None, 1)` but it is not

roberth (approved):
It seems to be complete now.
LGTM

PR #133, opened 2020-09-13
2 names and 1 ext added

lcnittl:


identify/extensions.py
```diff
@@ -183,11 +183,12 @@
     'vue': {'text', 'vue'},
     'war': {'binary', 'zip', 'jar'},
     'wav': {'binary', 'audio', 'wav'},
-    'wkt': {'text', 'wkt'},
     'whl': {'binary', 'wheel', 'zip'},
+    'wkt': {'text', 'wkt'},
     'woff': {'binary', 'woff'},
     'woff2': {'binary', 'woff2'},
     'wsgi': {'text', 'wsgi', 'python'},
+    'xhtml': {'text', 'xml', 'xhtml'},
```
lcnittl, line 191: Should `xhtml` have `html` type as well?

asottile:
can you sort this in a separate PR, (some of them probably shouldn't be sorted, for instance `adoc` and `asciidoc` are next to each other because the values are the same) -- as it is now, it's really difficult to see what was added in this PR

lcnittl:
Sure, done

lcnittl:
> for instance `adoc` and `asciidoc` are next to each other because the values are the same

Best use case for #134 :rofl:

identify/extensions.py
asottile: yeah probably, sorry missed this first time around!

asottile (approved):

identify/extensions.py
lcnittl: Sorry, saw it just now!

lcnittl: Shall I add it still in another PR?

asottile: up to you!

PR #136, opened 2020-09-16
Add 'html' type to 'xhtml'

lcnittl:


identify/extensions.py
```diff
@@ -241,6 +241,7 @@
     'CMakeLists.txt': EXTENSIONS['cmake'],
     'COPYING': EXTENSIONS['txt'],
     'Dockerfile': {'text', 'dockerfile'},
+    'flake8': EXTENSIONS['ini'] | {'flake8'},
```
asottile, line 244: a file named `flake8` isn't a valid flake8 configuration

lcnittl: Did I misinterpret [this section](https://flake8.pycqa.org/en/latest/user/configuration.html#user-configuration)?

```Note that Flake8 looks for ~\.flake8 on Windows and ~/.config/flake8 on Linux and other Unix systems.```

Or should it not be included as it is not a typical repo file?

asottile: yeah those aren't going to be checked in -- also I plan to remove the user configuration at some point -- it causes significant confusion and almost nobody uses it afaik

lcnittl: I see. I'll remove the commit and force push to my branch.

lcnittl: Or can you as collaborator drop the commit in the pull request?

asottile: I could, but it's easier if you do it!

lcnittl: Alright!

PR #137, opened 2020-09-18
Add '.mako' as an extension for mako templates

rouge8:
https://www.makotemplates.org/

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -95,6 +95,7 @@ EXTENSIONS = {
     'lidr': {'text', 'idris'},
     'lua': {'text', 'lua'},
     'm': {'text', 'c', 'objective-c'},
+    'mako': {'text', 'mako'},
     'manifest': {'text', 'manifest'},
     'map': {'text', 'map'},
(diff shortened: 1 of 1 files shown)


asottile:
this was considered in the past, but I don't believe mako actually uses `.mako` as a file extension (their examples are all `.txt` and `.html`) -- some more official documentation of the mako extension would be needed to accept this

rouge8:
Alembic (which is by the same author) uses `.mako` within its codebase :) https://github.com/sqlalchemy/alembic/tree/master/alembic/templates/generic

asottile:
I use .mako too, but it's not documented at all

rouge8:
Yeah, and it looks unlikely to be: https://github.com/sqlalchemy/mako/issues/50

GH Linguist uses `.mako` (or `.mao` which I’ve never seen anywhere) to identify Mako files: https://github.com/github/linguist/blob/8083cb5a89cee2d99f5a988f165994d0243f0d1e/lib/linguist/languages.yml#L3166. It’s also one of the extensions used by default in pyramid_mako, but yeah, nothing official.

PR #147, opened 2020-10-17
Identify can now detect python packages

saroad2:
This is useful in order to determine whether a directory should be evaluated or not.


asottile:
the only time directories will be passed to identify by pre-commit is when they are git submodules

asottile:
an `__init__.py` file also is not needed for a directory to be an importable python package:

```console
$ mkdir t
$ python -c 'import t'
$
```

you may want to use `pkgutil.walk_packages` -- but it also doesn't understand namespace packages

PR #150, opened 2020-11-02
add c# file types

rkm:


identify/extensions.py
```diff
@@ -27,6 +27,7 @@
     'cpp': {'text', 'c++'},
     'crt': {'text', 'pem'},
     'cs': {'text', 'c#'},
+    'csproj': {'text', 'c#'},
```
asottile, line 30: csproj isn't a c# source file, it should be its own type (it's xml as well right?) -- so probably `{'text', 'xml', 'csproj'}` -- same for `.sln`

rkm: Technically correct (the best kind of correct 😉 ). I've updated both, except that `sln` isn't an xml format.

Planning on using these with the new `dotnet` pre-commit hooks!

asottile (approved):

PR #152, opened 2020-11-03
Adds support for Arch Linux packaging description files

axel-kah:
Adds support for `PKGBUILD` and `*.install` files. Both are interpreted as bash by the Arch Linux packaging toolchain (`makepkg` and `pacman`) but do not use shebangs or common bash file extensions.

Reference: https://www.archlinux.org/pacman/PKGBUILD.5.html

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -257,6 +257,7 @@ NAMES = {
     'PATENTS': EXTENSIONS['txt'],
     'Pipfile': EXTENSIONS['toml'],
     'Pipfile.lock': EXTENSIONS['json'],
+    'PKGBUILD': {'text', 'bash', 'pkgbuild', 'alpm'},
     'README': EXTENSIONS['txt'],
     'Rakefile': EXTENSIONS['rb'],
     'setup.cfg': EXTENSIONS['ini'],


identify/extensions.py
```diff
@@ -73,6 +73,7 @@
     'idr': {'text', 'idris'},
     'inc': {'text', 'inc'},
     'ini': {'text', 'ini'},
+    'install': {'text', 'bash', 'alpm'},
```
asottile, line 76: some googling shows other things using ".install" as a file extension, I'm not sure this is unambiguously arch packaging -- thoughts?

axel-kah: Bummer. Seems some people use it for linux distributables in the widest sense. What might be enough is to factor in the executable bit. I checked and in the arch context this file does not have the executable bit set, but I can only assume that in a different context the bit is set.
If that is not good enough then I don't see a way around reading the file and determine the type from the function names used. In the arch context you have to define at least one function following the pattern `post_*`. Not sure how often you had to deal with this kind of ambiguity. It seems this case can't be solved by the current identify algorithm.

asottile: usually in ambiguity we lean on the side of caution and don't include the extension 🤷 

there intentionally aren't complicated discovery rules (bits set, file contents, magic numbers, etc.) so it probably means `.install` isn't doable :(

axel-kah: Too bad, but I get your point. Slightly off topic: do you know if pre-commit can be persuaded to feed this file type to the shellcheck hook for instance regardless of identify not classifying it as bash script?

asottile: you can do:

```yaml
-   id: shellcheck
-   id: shellcheck
    types: [file]
    files: \.install$
```

asottile (approved):

PR #155, opened 2020-11-22
Extend `js` and `ts` tags to include full names

boidolr:
JavaScript and TypeScript identification was mixed in the tags it assigned: either the extension or the full name.
I am adding both to make it more consistent and easier to use.


asottile:
#29 is going to be a better way to handle this

PR #156, opened 2020-11-22
Add support for Lektor extensions, additional tool config rcfiles and common plain-text meta-files lacking extensions

CAM-Gerlach:
* Add `.lr` files from the [Lektor](https://www.getlektor.com/) static CMS, and `.lektorproject` which == INI
* Add several config/rc files without extensions from additional relatively popular linting/etc. tools that have pre-commit hooks ([.browserslistrc](https://github.com/browserslist/browserslist#config-file), [.codespellrc](https://github.com/codespell-project/codespell#using-a-config-file), [.csslintrc](https://github.com/CSSLint/csslint/wiki/Command-line-interface#configuration-files), [.gitlint](https://jorisroovers.com/gitlint/configuration/#the-gitlint-file), [pylintrc](http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options)). I took care to only add such files where the un-extensioned config file had only one defined file type, as opposed to cases where it could potentially be in one of several
* Add a few additional common plain text meta filenames (CHANGELOG, CONTRIBUTING, [COPYING.LESSER](https://www.gnu.org/licenses/gpl-howto.html) NEWS) with relatively wide usage

There was one ambiguity here: Should ordering the `NAMES` dict be case-sensitive (i.e. should capitalized letters have higher precedence than lowercase)? Currently, this wasn't possible to infer from the existing order, as there were no cases I could find where this would have made a difference to ordering, but this PR adds one such instance (`pylintrc`). I arbitrarily picked non-case-sensitive ordering for now, but I'm happy to change it if desired, just lmk. Thanks!


identify/extensions.py
```diff
@@ -223,21 +225,26 @@
     '.bash_profile': EXTENSIONS['bash'],
     '.bashrc': EXTENSIONS['bash'],
     '.bowerrc': EXTENSIONS['json'] | {'bowerrc'},
+    '.browserslintrc': {'txt', 'browserslintrc'},
```
asottile, line 228: typo

identify/extensions.py
```diff
@@ -223,21 +225,26 @@
...
     '.gitignore': {'text', 'gitignore'},
+    '.gitlint': EXTENSIONS['ini'] | {'gitlint'},
     '.gitmodules': {'text', 'gitmodules'},
     '.hgrc': EXTENSIONS['ini'] | {'hgrc'},
     '.jshintrc': EXTENSIONS['json'] | {'jshintrc'},
     '.mailmap': {'text', 'mailmap'},
     '.mention-bot': EXTENSIONS['json'] | {'mention-bot'},
     '.npmignore': {'text', 'npmignore'},
     '.pdbrc': EXTENSIONS['py'] | {'pdbrc'},
+    '.pylintrc': EXTENSIONS['ini'] | {'pylintrc'},
```
asottile, line 247: I don't see this used in the documentation, the name there is `pylintrc` without a dot

identify/extensions.py
```diff
@@ -246,18 +253,24 @@
     'BUILD': {'text', 'bazel'},
     'BUILD.bazel': {'text', 'bazel'},
     'CMakeLists.txt': EXTENSIONS['cmake'],
+    'CHANGELOG': EXTENSIONS['txt'],
+    'CODE_OF_CONDUCT': EXTENSIONS['txt'],
+    'CONTRIBUTING': EXTENSIONS['txt'],
     'COPYING': EXTENSIONS['txt'],
     'Dockerfile': {'text', 'dockerfile'},
     'Gemfile': EXTENSIONS['rb'],
+    'GPL': EXTENSIONS['txt'],
```
asottile, line 262: is this actually used? I've never seen this

identify/extensions.py
```diff
@@ -246,18 +253,24 @@
     'BUILD': {'text', 'bazel'},
     'BUILD.bazel': {'text', 'bazel'},
     'CMakeLists.txt': EXTENSIONS['cmake'],
+    'CHANGELOG': EXTENSIONS['txt'],
+    'CODE_OF_CONDUCT': EXTENSIONS['txt'],
```
asottile, line 257: iirc github requires this to be `.md` or `.rst` -- is there conventions of this elsewhere?

asottile (commented):
- rebase
- fix the test that's failing
- provide links in the PR to the individual configuration formats

identify/extensions.py
CAM-Gerlach: Good catch, thanks. Fixed.

identify/extensions.py
CAM-Gerlach: Both are listed [here](http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options) in several places, and I've seen both used (unfortunately, I can't seem to search for one or the other separately on Github to evaluate actual numbers).

identify/extensions.py
CAM-Gerlach: Oops, I saw this in a project and misremembered it as the recommended practice for adding the LGPL, when the latter is in fact conventionally added as `COPYING.LESSER` per the FSF [GPL Howto](https://www.gnu.org/licenses/gpl-howto.html), so I substituted that instead.

identify/extensions.py
CAM-Gerlach: Ah, wasn't sure about that. Not that I could find, so I dropped it.

CAM-Gerlach:
> rebase

Done.

> fix the test that's failing

Not sure what you mean, all the tests were green before and are now?

> provide links in the PR to the individual configuration formats

Done, also updated for the changes you mentioned.

identify/extensions.py
```diff
@@ -223,21 +225,26 @@
     '.bash_profile': EXTENSIONS['bash'],
     '.bashrc': EXTENSIONS['bash'],
     '.bowerrc': EXTENSIONS['json'] | {'bowerrc'},
+    '.browserslintrc': {'text', 'browserslintrc'},
```
asottile, line 228: there's still a typo here

asottile:
> > fix the test that's failing
> 
> Not sure what you mean, all the tests were green before and are now?

I added a test on master -- you fixed it as part of fixing one of your two typos :)

identify/extensions.py
asottile: also is this file json?

asottile: ah no it's not, I read their docs wrong

CAM-Gerlach: Sorry about that, fixed the `txt` -> `text` one but not the name of the file. Fixed now, thanks.

identify/extensions.py
```diff
@@ -246,18 +253,23 @@
     'BUILD': {'text', 'bazel'},
     'BUILD.bazel': {'text', 'bazel'},
     'CMakeLists.txt': EXTENSIONS['cmake'],
+    'CHANGELOG': EXTENSIONS['txt'],
+    'CONTRIBUTING': EXTENSIONS['txt'],
     'COPYING': EXTENSIONS['txt'],
+    'COPYING.LESSER': EXTENSIONS['txt'],
```
asottile, line 259: iirc this isn't necessary -- it'll get covered by `COPYING`

identify/extensions.py
```diff
@@ -223,21 +225,26 @@
...
     '.gitignore': {'text', 'gitignore'},
+    '.gitlint': EXTENSIONS['ini'] | {'gitlint'},
     '.gitmodules': {'text', 'gitmodules'},
     '.hgrc': EXTENSIONS['ini'] | {'hgrc'},
     '.jshintrc': EXTENSIONS['json'] | {'jshintrc'},
     '.mailmap': {'text', 'mailmap'},
     '.mention-bot': EXTENSIONS['json'] | {'mention-bot'},
     '.npmignore': {'text', 'npmignore'},
     '.pdbrc': EXTENSIONS['py'] | {'pdbrc'},
+    '.pylintrc': EXTENSIONS['ini'] | {'pylintrc'},
```
asottile, line 247: iirc this isn't necessary if there's `pylintrc` -- can you double check this?

identify/extensions.py
CAM-Gerlach: Confirmed, thanks. Dropped that line.

identify/extensions.py
CAM-Gerlach: Also confirmed, dropped this as well.

asottile (approved):

CAM-Gerlach:
KITTEH :smile_cat:

asottile:
[miao miao](https://www.youtube.com/watch?v=2rDNJvHznrM)

CAM-Gerlach:
_Meows in Spanish_

Maybe a little out of scope, but why not defer the request until the click event fires for the approve radio button? Seems like it would save a lot of extra requests when doing loading any PR page without approving it. Too much latency needing to wait for the request after clicking approve, or just not optimized?

asottile:
too much latency, yeah

PR #165, opened 2021-01-19
Add audio file extensions

mondeja:
Add extensions for most common audio file types.


asottile:
these seem a little inconsistently tagged -- if we want a catchall `audio` there should also be a specific tag for each extension

also could you provide references for each extension?  it might be easier to review if split up as well

mondeja:
> if we want a catchall audio there should also be a specific tag for each extension

Do you mean something like, for example, `windows-media-audio` for `wma`?

> also could you provide references for each extension?

All are listed in [Audio File Format (Wikipedia)](https://en.wikipedia.org/wiki/Audio_file_format). Let me know if you need more information for any of them.

asottile:
well for example you have `mp3: {'binary', 'mp3'}` which isn't tagged `'audio'` and you have `aa` which doesn't have any specific tagging (what is `aa`?)

I don't think that `audio` is a useful tag either, I'm not sure what tool could support every extension of the same name (and tagging `binary` doesn't really help either since that can be detected).  The only benefit to tagging these I'd imagine is some disk io to avoid the binary check, I don't see how a tool would utilize these tags

perhaps you could provide an example?

> All are listed in Audio File Format (Wikipedia). Let me know if you need more information for any of them.

I'm sorry, I don't have time to read through a wikipedia article -- please provide some digestable citations for each addition

mondeja:
> perhaps you could provide an example?

I'm trying to resolve a simple problem, could a file be *in normal conditions* an audio file without open it? The [`sndhdr`](https://docs.python.org/3/library/sndhdr.html) standard module doesn't help here because opens the file and doesn't supports formats widely used as MP3. I'm currently using [`mimetypes`](https://docs.python.org/3/library/mimetypes.html), but doesn't supports weird file types and relies in platform-dependent files. I'm facing a situation where an user can pass an audio format to a video-only parameter, so I want to tell the user, "hey this is audio, not video" after a reading error, so I can improve the API of a library.

I understand if this is not within the scope of the library, just tell me if is the case before post here the full list of references.

mondeja:
- `.aa` and `.aax`: https://en.wikipedia.org/wiki/Audible_(service)#Quality
- `.aac`: https://en.wikipedia.org/wiki/Advanced_Audio_Coding
- `.act`: https://en.wikipedia.org/wiki/ACT_(audio_format)
- `.aiff`: https://en.wikipedia.org/wiki/Audio_Interchange_File_Format
- `.alac`: https://en.wikipedia.org/wiki/Apple_Lossless
- `.au`: https://en.wikipedia.org/wiki/Au_file_format
- `.flac`: https://en.wikipedia.org/wiki/FLAC
- `.gsm`: https://en.wikipedia.org/wiki/Full_Rate
- `.m4a`: Audio data stream of https://en.wikipedia.org/wiki/MPEG-4_Part_14
- `.mp3`: https://en.wikipedia.org/wiki/MP3
- `.ogg`: https://en.wikipedia.org/wiki/Ogg (maybe adding "multimedia" tag apart from "audio", "video"?)
- `.rf64`: https://en.wikipedia.org/wiki/RF64
- `.wma`: https://en.wikipedia.org/wiki/Windows_Media_Audio
- `.wv`: https://en.wikipedia.org/wiki/WavPack

asottile:
Ideally we'd only have extensions supported that people actually use, I don't see a need for a lot of the ones you've added here.  Also each of them should have a specific tag as well not just "audio", otherwise a tool written specifically for one extension cannot possibly target that extension using tags.  For example, if I wrote a tool targetting just mp3 files, I would want to be able to use `types: [mp3]` but I cannot with this patch as written

mondeja:
> I don't see a need for a lot of the ones you've added here.

Okay, which ones do you don't want here? Let me know, I remove then and wil add tags for the rest.

mondeja:
By the way, I'm not sure about the description of this library, it says "file identification library for Python", but after your suggestion perhaps should be "file identification library for pre-commit" not? It seems that you don't want to accept some file types for some reason. It's fine, just say it and I will write my own "Audio file identification library for Python".

asottile:
> > I don't see a need for a lot of the ones you've added here.
> 
> Okay, which ones do you don't want here? I'm not sure of what people use, except `.mp3`, `.flac` and perhaps `.wma` which seems to me that are the most used in this list. Let me know, I remove then and will add tags for the rest.

I'd probably keep these ones, these are the only ones I've heard of that are in your pull request currently: aac, flac, m4a, mp3, ogg, wma

> By the way, I'm not sure about the description of this library, it says "file identification library for Python", but after your suggestion perhaps should be "file identification library for pre-commit", not? It seems that you don't want to accept some file types for some reason. It's fine, just say it and I will write my own "Audio file identification library for Python".

It's not specifically for pre-commit (though pre-commit is the largest consumer), but we have to strike a balance between accepting things and not.  If (for example) some obscure file type is added that also has a less obscure usage we'd be a bit painted into a corner in making a breaking change to adjust that.  As such, I think it really only makes sense to include common file types and not obscure or obsolete ones (especially if you have no use for the less common ones).  The mimetypes library has a similar mantra in that it doesn't include some of the very obscure ones you've added here but I don't think you'd make the same assertion about it being wrong :)

mondeja:
In that case, I think that the most reasonable is to make my own library focused in audio file types only.
Anyway, thank you for the feedback :pray:

PR #169, opened 2021-02-25
Add global constants to classify tags by type

### LINKED ISSUE #157: Provide a way to retrieve tags structured by tag type
Currently, `identify` only outputs tags as an unordered set, without any further semantics to allow a consumer to determine which tags correspond to which properties (object type, binary-ness, executable-ness, and file format). For consumption by other tools, it would be very useful to be able to consider each "type" of tag separately; otherwise, to determine the tags in each categories, consumers would need to resort to brittle hardcoding of the which individual tags belong in each category.

Therefore, it would be very useful if there was a means for identify to output its tags structured by type, perhaps as a dictionary with the categories as keys (e.g. `type`, `mode`, `binary`, `format`) as keys and a set of the corresponding tag(s) as the values for each.

To implement this with minimal churn and a backward-compatible public API, the code in `tags_from_path()` could be refactored into another public function, e.g. `get_structured_tags_from_filename()` and add the resulting tags to the set corresponding to the property in question (with an extra line to handle peeling off text and binary into a separate set from the file format parameters, to avoid having to refactor the entirety of `extensions.py`). This function could be used by other packages needing the tags in a structured form, as well as by`tags_from_filename` for concatenation into a single set of tags as current consumers of that function expect.

For some context, I'm in the early stages of building a lightweight open-source tool integrating with pre-commit for managing linter, fixer, checker, and formatter config, and I'm looking to have per-file-type settings (e.g. indentation) inherit properly while using identify's existing standardized tags, i.e. the user sets default settings (indentation, tools to enable, etc) `file`, which can be overriden by `text`, and overriden in turn by e.g. `python` (this particular order being a concern of the consumer, not identify). I could hardcode the tags that belong to each property (and assume the rest are formats), but it would be much more robust and useful if Identify could provide this structured information itself.

I can submit a PR to implement this plus some simple tests, as it should be fairly straightforward to implement, and am also happy to hear further suggestions/recommendations with regard to implementation, how you would suggest structuring the output, etc. Thanks!

CAM-Gerlach:
Per the (lengthy) discussion in #157 , implements the proposed global constants to classify tags by type, making the code and tests slightly simpler and more robust, and enabling external tools consuming them to differentiate between tag types without resorting to hardcoding.

Also adds a test that all tags from each group are included in the list of all tags, and expands the existing basic tags test to at least one tag from each group.

I'm happy to change the names chosen for the constants if the currently selected ones are problematic, just let me know, thanks!

Resolves #157

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -25,7 +25,10 @@ NON_EXECUTABLE = 'non-executable'
 TEXT = 'text'
 BINARY = 'binary'
 
-ALL_TAGS = {DIRECTORY, SYMLINK, FILE, EXECUTABLE, NON_EXECUTABLE, TEXT, BINARY}
+TYPE_TAGS = frozenset((DIRECTORY, FILE, SYMLINK))
+MODE_TAGS = frozenset((EXECUTABLE, NON_EXECUTABLE))
+ENCODING_TAGS = frozenset((BINARY, TEXT))
+ALL_TAGS = {*TYPE_TAGS, *MODE_TAGS, *ENCODING_TAGS}
 ALL_TAGS.update(*extensions.EXTENSIONS.values())
 ALL_TAGS.update(*extensions.EXTENSIONS_NEED_BINARY_CHECK.values())
 ALL_TAGS.update(*extensions.NAMES.values())
@@ -62,14 +65,14 @@ def tags_from_path(path):
 
     # some extensions can be both binary and text
     # see EXTENSIONS_NEED_BINARY_CHECK
-    if not {TEXT, BINARY} & tags:
+    if not ENCODING_TAGS & tags:
         if file_is_text(path):
             tags.add(TEXT)
         else:
             tags.add(BINARY)
 
-    assert {TEXT, BINARY} & tags, tags
-    assert {EXECUTABLE, NON_EXECUTABLE} & tags, tags
+    assert ENCODING_TAGS & tags, tags
+    assert MODE_TAGS & tags, tags
     return tags
 
 
--- a/tests/identify_test.py
+++ b/tests/identify_test.py
@@ -14,6 +14,20 @@ from identify import identify
 def test_all_tags_includes_basic_ones():
     assert 'file' in identify.ALL_TAGS
     assert 'directory' in identify.ALL_TAGS
+    assert 'executable' in identify.ALL_TAGS
+    assert 'text' in identify.ALL_TAGS
+
+
+@pytest.mark.parametrize(
+    'tag_group',
+    (
+        identify.TYPE_TAGS,
+        identify.MODE_TAGS,
+        identify.ENCODING_TAGS,
+    ),
+)
+def test_all_tags_contains_all_groups(tag_group):
+    assert tag_group < identify.ALL_TAGS
 
 
 def test_all_tags_contains_each_type():


identify/identify.py
```diff
@@ -25,12 +25,16 @@
...
+TYPE_TAGS = frozenset({DIRECTORY, FILE, SYMLINK})
+MODE_TAGS = frozenset({EXECUTABLE, NON_EXECUTABLE})
+ENCODING_TAGS = frozenset({BINARY, TEXT})
+FORMAT_TAGS = frozenset().union(
+    *extensions.EXTENSIONS.values(),
+    *extensions.EXTENSIONS_NEED_BINARY_CHECK.values(),
+    *extensions.NAMES.values(),
+    *interpreters.INTERPRETERS.values(),
+) - ENCODING_TAGS
+ALL_TAGS = frozenset({*TYPE_TAGS, *MODE_TAGS, *ENCODING_TAGS, *FORMAT_TAGS})
```
asottile, line 37: this does a lot of copies that weren't present in the left hand side -- can you validate that it doesn't make import time significantly slower for this module?

asottile: actually, this can be done much simpler by just adding your new constants after the old code instead of modifying how the current set of `ALL_TAGS` is built, then unbuilt, then built

CAM-Gerlach: Running a simple script that tests the time to import `identify.identify` with `python <testscript.py> in a conda environment with Python 3.7.9 and the respective dev versions of identify isntalled with `pip install -e .` over 12 runs, discarding the first two:

Import time, this PR: 30.6 ms
Import time, `master`: 30.9 ms

A small advantage to this PR, but well within the variability of the test runs.

Measuring the `time.time()` elapsed running the actual code block modified in `identify.identify`, averaged over 1000 runs in a loop:

Prior to PR: 39.4 us
With PR: 61.4 us

The same, but placed in a function and measured with IPython `%timeit`:

Prior to PR: 41.8 us +/- 2.08 us
With PR: 63.3 us +/- 1.18 us

Test script:

```python
import time
t1 = time.time()
import identify.identify
t2 = time.time()
delta_t = t2 - t1
print(delta_t)
```

CAM-Gerlach: To give some fun context to the above results, assuming the ≈20 us extra time cost per import suggested above, it will take around 100 million Identify imports to equal the ≈30 minutes it took me to set up, execute, tabulate and write up the above tests.

asottile: you're not really profiling properly, imports only execute the first time they're run

but also it isn't what we agreed upon in the original issue -- adding constants, not rewriting the current code (as much as I know you love rewriting things!)

CAM-Gerlach: Sorry, for some reason Github wasn't showing your other comments on this PR until I force-reloaded.

> actually, this can be done much simpler by just adding your new constants after the old code instead of modifying how the current set of `ALL_TAGS` is built, then unbuilt, then built

> but also it isn't what we agreed upon in the original issue -- adding constants, not rewriting the current code

Actually, I originally implemented this using the exact code snippit I'd proposed in #157 (aside from bug fixes) that I thought was what you had said you were okay with. Regardless, your suggestion makes much more sense, and I went back and restored the previous `ALL_TAGS` building process, just adding the hardcoded tag group constants before and the format tag group after. As you expected, this eliminates most of the small performance hit from this PR, with the code block now taking only 48.3 us to execute, vs. 61.4 us before, only 6.5 us slower than the previous 41.8 us, and nearly 4 OoM less than the import time (30 ms).

Also, I've rebased to separate each atomic change into a separate commit for easier review, and so I can drop any (e.g. change to use the constants rather than hardcoding in the tests) if not desired.

> you're not really profiling properly, imports only execute the first time they're run

As described above, the benchmarks reported involve either starting a fresh interpreter every iteration, or measuring the execution time of just running the specific code block in question (since run to run variability and the precision of my machine's OS timers for the former greatly outweighs the latter performance impact). Neither is subject to the import system's caching of previously loaded modules; I'm not importing in a for loop or in `%timeit` inside the same interpreter or anything silly like that, which would result in a time >4 OoM smaller than reported above (on the order of 1 us, not 30 ms). If there's some other way you'd like me to profile it that I haven't already, please let me know.

> (as much as I know you love rewriting things!)

Yeah, most scientists don't spend enough time refactoring their code, but I think sometimes I take it a _little_ too far in the opposite direction...

CAM-Gerlach:
@asottile I assume you dropped the `FORMAT_TAGS` constant due to the (extremely small but non-zero) runtime cost of generating it, given that consuming code can derive it from the existing constants (under assumptions regarding the other tag types that are likely to be invariant over the foreseeable future)?

asottile:
> @asottile I assume you dropped the `FORMAT_TAGS` constant due to the (extremely small but non-zero) runtime cost of generating it, given that consuming code can derive it from the existing constants (under assumptions regarding the other tag types that are likely to be invariant over the foreseeable future)?

I dropped it because it's unnecessary

CAM-Gerlach:
Right, for the reason mentioned above?

CAM-Gerlach:
Also, where's my cat? You know the only reason I do these PRs is for the adorable cat GIFs, right? :laughing:

PR #170, opened 2021-02-27
Add `avif` file extension

boidolr:
Add `.avif` extension to recognize files from extensions.


asottile:
please provide documentation for this extension and intended usecase

boidolr:
Sure. This is for the https://en.wikipedia.org/wiki/AV1#AV1_Image_File_Format_(AVIF) image format. I want to trigger a hook which uses pillow & pillow-avif-plugin to check for dimensions, and other properties of the file.

asottile (approved):

Issue #72, opened 2019-03-16
Identify named pipes and sockets


mitzkia:
The current version (`v1.4.0`) can not identify named pipes and socket files. 
I do not know if identify should work on this type of files, but I can show the reproduction, and can add a suggestion for fixing it.

Reproduction for named-pipes (in the last command it hangs):
```console
$ mkfifo /tmp/my-custom-named-pipe
$ file /tmp/my-custom-named-pipe
/tmp/my-custom-named-pipe: fifo (named pipe)
$ identify-cli /tmp/my-custom-named-pipe
```

bt:
```
^CTraceback (most recent call last):
  File "/usr/local/bin/identify-cli", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/identify/cli.py", line 23, in main
    tags = sorted(func(args.path))
  File "/usr/local/lib/python3.6/dist-packages/identify/identify.py", line 66, in tags_from_path
    if file_is_text(path):
  File "/usr/local/lib/python3.6/dist-packages/identify/identify.py", line 128, in file_is_text
    with open(path, 'rb') as f:
KeyboardInterrupt
```

Reproduction for socket files:
```console
$ python -c "import socket as s; sock = s.socket(s.AF_UNIX); sock.bind('/tmp/my-custom-socket')"
$ file /tmp/my-custom-socket 
/tmp/my-custom-socket: socket
$ identify-cli /tmp/my-custom-named-pipe
```

bt:
```
Traceback (most recent call last):
  File "/usr/local/bin/identify-cli", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/identify/cli.py", line 23, in main
    tags = sorted(func(args.path))
  File "/usr/local/lib/python3.6/dist-packages/identify/identify.py", line 59, in tags_from_path
    shebang = parse_shebang_from_file(path)
  File "/usr/local/lib/python3.6/dist-packages/identify/identify.py", line 172, in parse_shebang_from_file
    with open(path, 'rb') as f:
OSError: [Errno 6] No such device or address: '/tmp/my-custom-socket'
```

My suggestion is to fix these issues is to use `stat` module of python. With the help of this modul identify could check all other file types as well. 
URL: https://docs.python.org/3/library/stat.html

If possible I can propose a PR with pleasure.

asottile:
sounds fine to me! would you like to propose a PR?

PR #177, opened 2021-03-02
Identify sockets

mxr:
This is part of #72

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -1,6 +1,7 @@
 import os.path
 import re
 import shlex
+import stat
 import string
 import sys
 from typing import IO
@@ -18,13 +19,14 @@ printable = frozenset(string.printable)
 
 DIRECTORY = 'directory'
 SYMLINK = 'symlink'
+SOCKET = 'socket'
 FILE = 'file'
 EXECUTABLE = 'executable'
 NON_EXECUTABLE = 'non-executable'
 TEXT = 'text'
 BINARY = 'binary'
 
-TYPE_TAGS = frozenset((DIRECTORY, FILE, SYMLINK))
+TYPE_TAGS = frozenset((DIRECTORY, FILE, SYMLINK, SOCKET))
 MODE_TAGS = frozenset((EXECUTABLE, NON_EXECUTABLE))
 ENCODING_TAGS = frozenset((BINARY, TEXT))
 _ALL_TAGS = {*TYPE_TAGS, *MODE_TAGS, *ENCODING_TAGS}
@@ -36,12 +38,18 @@ ALL_TAGS = frozenset(_ALL_TAGS)
 
 
 def tags_from_path(path: str) -> Set[str]:
-    if not os.path.lexists(path):
+    try:
+        sr = os.lstat(path)
+    except (OSError, ValueError):  # same error-handling as `os.lexists()`
         raise ValueError(f'{path} does not exist.')
-    if os.path.isdir(path):
+
+    mode = sr.st_mode
+    if stat.S_ISDIR(mode):
         return {DIRECTORY}
-    if os.path.islink(path):
+    if stat.S_ISLNK(mode):
         return {SYMLINK}
+    if stat.S_ISSOCK(mode):
+        return {SOCKET}
 
     tags = {FILE}
 
--- a/tests/identify_test.py
+++ b/tests/identify_test.py
@@ -1,6 +1,8 @@
 import io
 import os
+import socket
 import stat
+from tempfile import TemporaryDirectory
 
 import pytest
 
@@ -12,6 +14,7 @@ def test_all_tags_includes_basic_ones():
     assert 'directory' in identify.ALL_TAGS
     assert 'executable' in identify.ALL_TAGS
     assert 'text' in identify.ALL_TAGS
+    assert 'socket' in identify.ALL_TAGS
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


tests/identify_test.py
```diff
@@ -51,6 +53,18 @@ def test_tags_from_path_symlink(tmpdir):
     assert identify.tags_from_path(x.strpath) == {'symlink'}
 
 
+def test_tags_from_path_socket(tmpdir):
+    cwd = os.getcwd()
```
asottile, line 57: is the chdir necessary? -- should be able to just do:

```python
socket_path = tmpdir.join('socket')
sock = socket.socket(socket.AF_UNIX)
sock.bind(str(socket_path))
# TODO: probably need cleanup code of the socket

assert identify.tags_from_path(socket_path) == {'socket'}

Comment (mxr) [reply tests/identify_test.py:57]: My first attempt was like that but I got an error that the path was too long. From some cursory Googling it looks like there is a hardcoded limit that I can't really change, but using a relative path worked

```console
$ pytest --pdb -k socket tests/identify_test.py
======================================================================================================================= test session starts =======================================================================================================================
platform darwin -- Python 3.8.2, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /Users/mxr/src/go/src/github.com/mxr/identify
plugins: cov-2.11.1
collected 83 items / 82 deselected / 1 selected                                                                                                                                                                                                                   

tests/identify_test.py F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

tmpdir = local('/private/var/folders/3l/njs4g0nx77v3z1d5085tqtlh0000gp/T/pytest-of-mxr/pytest-24/test_tags_from_path_socket0')

    def test_tags_from_path_socket(tmpdir):
        socket_path = tmpdir.join('socket')
        sock = socket.socket(socket.AF_UNIX)
>       sock.bind(str(socket_path))
E       OSError: AF_UNIX path too long

tests/identify_test.py:59: OSError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/mxr/src/go/src/github.com/mxr/identify/tests/identify_test.py(59)test_tags_from_path_socket()
-> sock.bind(str(socket_path))
(Pdb) socket_path
local('/private/var/folders/3l/njs4g0nx77v3z1d5085tqtlh0000gp/T/pytest-of-mxr/pytest-24/test_tags_from_path_socket0/socket')

```

asottile: changing directories shouldn't make a difference there -- the socket'll still have the same length (though global iirc if you don't specify an absolute path)

if you use `TemporaryDirectory(dir='/tmp')` does it improve the problem? -- macos is kinda bad about having short tempdirs

mxr: > changing directories shouldn't make a difference there -- the socket'll still have the same length (though global iirc if you don't specify an absolute path)

Maybe what's happening is socket creation succeeds (evidenced in the test) but subsequent use of the socket will fail in other ways?

> if you use `TemporaryDirectory(dir='/tmp')` does it improve the problem? -- macos is kinda bad about having short tempdirs

Sure I'll try that instead

identify/identify.py
```diff
@@ -42,6 +44,8 @@ def tags_from_path(path: str) -> Set[str]:
         return {DIRECTORY}
     if os.path.islink(path):
         return {SYMLINK}
+    if stat.S_ISSOCK(os.stat(path).st_mode):
+        return {SOCKET}
```
asottile, line 48: I wonder if we should do a single `os.stat` call and then decide based on that -- as it is right now this is going to do at least 4 stat calls to determine these tags

mxr: Do you mean `isdir`, `islink`, and others do `os.stat` under the hood?

asottile: yep, and lexists

mxr: For `os.access` do you recommend hand-rolling something that supports both posix & windows or leaving it as-is? Also for some of the public functions that check if a file exists & other properties, should I remove those checks (because the caller checks them) to save on stat calls, or preserve the behavior because they're public?

mxr: Went with erring towards leaving more code as-is but lmk if there's anything else straightforward that I missed

asottile (approved):

PR #178, opened 2021-03-02
Run CI on py39

mxr:


asottile:
unless there's any specific breaking let's hold off on this -- I'll do it automatically at some point

mxr:
Sure, how can I find out more about stability of py39 or in other words develop enough context so I don't open this PR in the first place 😆

asottile:
identify currently doesn't have any version-specific or version-sensitive code so it's very unlikely to be broken by a new python version -- you can run `tox -e py39` locally

PR #181, opened 2021-03-13
List Gemfile.lock as type 'text'

sirosen:
To avoid it being treated as ruby code (e.g. via pre-commit running tools like rubocop).

---

I'm not super confident about this as I don't work much in ruby, but my understanding of `Gemfile.lock` is that it's its own format and not ruby code (vs, for example, `Gemfile`). I started looking at `identify` because my pre-commit config is running rubocop on `Gemfile.lock` and blowing up.

I confirmed with `identify-cli` that this removes the ruby tag.

I'm guessing that somehow the content-based detection is treating my `Gemfile.lock` as ruby? I don't know what the explanation could be other than that.

If I'm wrong about how to handle this, please educate me! 😄

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -261,6 +261,7 @@ NAMES = {
     'COPYING': EXTENSIONS['txt'],
     'Dockerfile': {'text', 'dockerfile'},
     'Gemfile': EXTENSIONS['rb'],
+    'Gemfile.lock': {'text'},
     'Jenkinsfile': {'text', 'groovy'},
     'LICENSE': EXTENSIONS['txt'],
     'MAINTAINERS': EXTENSIONS['txt'],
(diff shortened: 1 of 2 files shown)


asottile:
can you add a quick testcase demonstrating this -- want to make sure it doesn't regress in the future (especially because this is kind of a special special case).

sirosen:
I just added `Gemfile` and `Gemfile.lock` to the filename -> tags test.

asottile (approved):
(just a small tip for next time -- when making a pull request do it from a feature branch and not from `master` / `main`)

sirosen:
I see you already did a release. The way you run these projects makes it a huge pleasure to contribute!

I will not be lazy and actually make a feature branch next time.

PR #188, opened 2021-03-27
Add .dll as a binary file

kaidokert:
I have a few binary test vectors that pre-commit wants to mess up, calling those .bin seems like pretty solid solution. 

Also, .exe exists, .dll is a very close cousin

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -32,6 +32,7 @@ EXTENSIONS = {
     'cxx': {'text', 'c++'},
     'dart': {'text', 'dart'},
     'def': {'text', 'def'},
+    'dll': {'binary'},
     'dtd': {'text', 'dtd'},
     'ear': {'binary', 'zip', 'jar'},
     'edn': {'text', 'clojure', 'edn'},


identify/extensions.py
```diff
@@ -32,6 +33,7 @@
     'cxx': {'text', 'c++'},
     'dart': {'text', 'dart'},
     'def': {'text', 'def'},
+    'dll': {'binary'},
```
asottile, line 36: I'm fine adding `dll`, but `.bin` is too generic of an extension (and doesn't necessarily mean it contains binary data)

you can also use `exclude: \.bin$` to exclude it.  if identify is identifying this as text it's likely that it doesn't contain any unprintable bytes?

kaidokert: They do actually contain printable bytes, but are still to be treated as binary, being specifically bit test vectors for binary protocols. ( i.e. end-of-line-fixer and a number of other generic pre-commit hooks shouldn't touch these )

Adding an exclude to every hook is a workaround, but does make the config file quite a bit more verbose

asottile: do you know about [global exclude](https://pre-commit.com/#top_level-exclude) ?

kaidokert: I do, but there are other hooks that do need to run on those, so either way there's extra configuration that could be solved by convention.

asottile (approved):

PR #189, opened 2021-03-28
fix: checking for executable file regardless containerization

geeky-smurf:
- relates to pre-commit/pre-commit-hooks#528

for more information, see docker/for-mac#5029

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -1,5 +1,5 @@
 import errno
-import os.path
+import os
 import re
 import shlex
 import stat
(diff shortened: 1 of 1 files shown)


asottile:
nah sorry, docker-for-mac should fix their bug

geeky-smurf:
@asottile: docker/for-mac#5029 is open since Oct 2020 and will probably will not fix. why not improving identify implementation to work on every nix like?

asottile:
I'm not about to change my software because docker inc decided they can ship buggy software, sorry

PR #195, opened 2021-04-20
Detect .cuh as CUDA header file.

bdice:
Hi @asottile and collaborators! I would like to add support for CUDA header files, which often have the extension `.cuh`. This [StackOverflow post](https://stackoverflow.com/a/2393855) can be used as supporting evidence. Thank you, pre-commit is a wonderful tool.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -29,6 +29,7 @@ EXTENSIONS = {
     'css': {'text', 'css'},
     'csv': {'text', 'csv'},
     'cu': {'text', 'cuda'},
+    'cuh': {'text', 'header', 'cuda'},
     'cxx': {'text', 'c++'},
     'dart': {'text', 'dart'},
     'def': {'text', 'def'},
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -29,7 +29,7 @@ EXTENSIONS = {
     'css': {'text', 'css'},
     'csv': {'text', 'csv'},
     'cu': {'text', 'cuda'},
-    'cuh': {'text', 'header', 'cuda'},
+    'cuh': {'text', 'cuda'},
     'cxx': {'text', 'c++'},
     'dart': {'text', 'dart'},
     'def': {'text', 'def'},


identify/extensions.py
```diff
@@ -29,6 +29,7 @@
     'css': {'text', 'css'},
     'csv': {'text', 'csv'},
     'cu': {'text', 'cuda'},
+    'cuh': {'text', 'header', 'cuda'},
```
asottile, line 32: just a few questions (also I would probably want a stronger source than a SO post -- perhaps something official from the CUDA project?)

- can all `.cuh` files be interpreted as C headers? otherwise the `'header'` tag might not be appropriate here
- can all `.cuh` files be interpreted the same as `.cu` files would be?

bdice: @asottile Thanks for the review!

There are official CUDA sample codes and popular CUDA applications that use `.cuh`, here is a [specific example](https://github.com/NVIDIA/cuda-samples/tree/568b39bd5bea5924446d49d07a4c026a1ec945a5/Samples/shfl_scan) and [a longer list](https://github.com/NVIDIA/cuda-samples/search?q=cuh), as well as some NVIDIA provided libraries like [cub](https://github.com/NVIDIA/cub/tree/main/cub) and non-NVIDIA applications like [HOOMD-blue](https://github.com/glotzerlab/hoomd-blue/tree/master/hoomd).

- `.cuh` is used for CUDA C/C++ headers, so they're not plain C/C++. We can remove `'header'` if that's the intended distinction, I am just trying to follow the right pattern. 😄
- I'm not 100% sure what you're asking in the second question, but the general idea is that `.cuh` is to `.h` as `.cu` is to `.cc` or `.cpp`. A `.cuh` file is a header file included by a `.cu` file that may include CUDA C/C++ keywords and syntax.

I would be fine with this suggestion if it makes more sense to you.
```suggestion
    'cuh': {'text', 'cuda'},
```

asottile: yeah that seems fine !

bdice: Cool, I fixed that. 👍

PR #199, opened 2021-05-05
Add unity file extensions

JensHeinrich:
This adds the unity specific file extensions which are mostly tank files


identify/extensions.py
```diff
@@ -1,6 +1,9 @@
 EXTENSIONS = {
     'adoc': {'text', 'asciidoc'},
+    'anim': {'text', 'yaml'},
     'asciidoc': {'text', 'asciidoc'},
+    'asmdef': {'text', 'json'},
+    'asset': {'text', 'yaml'},
```
asottile, line 6: this, along with several of the other names are too generic to be accepted -- for the others please provide references in the PR description

JensHeinrich: Would a peaking guesser like the one for shebangs acceptable for yaml?
Because that are the extensions unity3d uses for its internal yaml stuff

asottile: no, sorry

asottile: note that you can always filter via `files: ...` for more specialized cases

PR #201, opened 2021-05-12
propose adding Qt `ui` file as xml

davidhyman:
Some projects using the Qt framework rely on `.xml` files that define UI layouts.
More [info on the format here](https://doc.qt.io/qtforpython/tutorials/basictutorial/uifiles.html).

On the topic of Qt, there's also [`.qrc`](https://doc.qt.io/qt-5/resources.html) and possibly others, probably for a different MR though.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -185,6 +185,7 @@ EXTENSIONS = {
     'twig': {'text', 'twig'},
     'txsprofile': {'text', 'ini', 'txsprofile'},
     'txt': {'text', 'plain-text'},
+    'ui': {'text', 'xml'},
     'v': {'text', 'verilog'},
     'vb': {'text', 'vb'},
     'vbproj': {'text', 'xml', 'vbproj'},


identify/extensions.py
```diff
@@ -185,6 +185,7 @@
     'twig': {'text', 'twig'},
     'txsprofile': {'text', 'ini', 'txsprofile'},
     'txt': {'text', 'plain-text'},
+    'ui': {'text', 'xml'},
```
asottile, line 188: unfortunately this file extension does not unambiguously belong to qt so I can't accept this PR

davidhyman: @asottile thanks for taking a look, but also disappointed to see this closed so quicky.

- Which other projects/file-contents are you referring to that cause ambiguity?
- Do you have some guidelines as to how unambiguous an extension needs to be for inclusion?
  - It was my understanding that all file extensions are conventions anyway, so always include an element of ambiguity, and presumably `identify` is constrained to that same limitation?
  - I am curious about what the threshold is for causing a false positive on an obscure filetype vs having to 'escape' the library for more common files?
- In order to support the functionality of detecting qt `ui` files, would there be an alternative implementation, using `identify`, that you could approve of?
  - e.g. [well-formed xml should begin with an XML declaration](https://www.w3.org/TR/xml/#sec-prolog-dtd) which could be detected as part of [reading the first 1024 bytes](https://github.com/pre-commit/identify/blob/master/identify/identify.py#L133)?

asottile: `ui` is a generic word, it can't possible be unambiguous in this way -- if it were `.qtui` it would be easily unambiguous

reading files is not an option, `identify` is intentionally simple so it can be fast and does not aim to have full coverage of all file types (which is impossible)

you can always use `files: \.ui$` yourself if you want detection for that filetype

PR #202, opened 2021-05-14
Add .jenkins extension type

jhancock93:
Creates a .jenkins extension and filetype. Detects 'Jenkinsfile' and *.jenkins as that type.


identify/extensions.py
```diff
@@ -247,6 +247,7 @@
     '.gitlint': EXTENSIONS['ini'] | {'gitlint'},
     '.gitmodules': {'text', 'gitmodules'},
     '.hgrc': EXTENSIONS['ini'] | {'hgrc'},
+    '.jenkins': EXTENSIONS['jenkins'],
```
asottile, line 250: do people make jenkinsfiles that are just `.jenkins` -- I don't think this is necessary due to the extensions listed above -- I could be wrong here though (?)

jhancock93: There are times where a repository might have multiple jenkins files. One for CI/CD purposes and one for deployment purposes. They can't all have the name jenkinsfile, in which case a .jenkins extension is a good convention for other pipelines.

asottile: what I'm saying is this entry right here is redundant **unless** the filename is exactly `.jenkins` (with no name, just extension)

asottile (approved):

PR #208, opened 2021-06-01
bats file extension added

leventyalcin:
bats is a TAP-compliant testing framework [1]. However, .bats file are bash and shellcheck compatible.

Recognising this files as bash will help them to test with shellcheck

https://github.com/bats-core/

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -6,6 +6,7 @@ EXTENSIONS = {
     'avif': {'binary', 'image', 'avif'},
     'bash': {'text', 'shell', 'bash'},
     'bat': {'text', 'batch'},
+    'bats': {'text', 'shell', 'bash', 'bats'},
     'bib': {'text', 'bib'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -6,6 +6,7 @@
     'avif': {'binary', 'image', 'avif'},
     'bash': {'text', 'shell', 'bash'},
     'bat': {'text', 'batch'},
+    'bats': {'text', 'shell', 'bash'},
```
asottile, line 9: this should probably have `bats` as well so that if bats-specific linting is needed it can be done against that `type`

leventyalcin: I hope I'm getting this correctly. 

I didn't add `bats` as a `type` due to the fact that there is no `bats` specific linter etc but everything is based on bash. However, there can good use cases in the future. So, the next commit incudes that, too.

asottile (approved):

PR #209, opened 2021-06-02
Added Zope pt, zpt and zcml

gotcha:


identify/extensions.py
```diff
@@ -135,6 +135,7 @@
     'pp': {'text', 'puppet'},
     'properties': {'text', 'java-properties'},
     'proto': {'text', 'proto'},
+    'pt': {'text', 'zpt'},
```
asottile, line 138: pt is likely too generic, a quick google search does not turn up zope but some other thing

gotcha: Trying to understand what you mean: do you mean that `zpt` is too generic ? or that the `pt` extension is used by too many apps ?

If the latter, how does identification works when a file extension is used by different systems ?

asottile: the latter

identification does not work in that case, we rejected extensions which are too uncommon or unofficial or ambiguous

gotcha: I removed the `pt` registration.

As you might guess, my additions were made in the context of `pre-commit`.

Is it correct that to specify hooks for Zope page templates, usage of the `files` setting will then remain compulsory for `pt` files ? Or is there a better way ?

asottile: yep, `files` will be necessary for that -- to be honest, `types` doesn't buy you much over `files` except that you avoid the regex (there's no shebangs involved which is where it really benefits)

asottile (approved):

PR #232, opened 2021-09-17
add hcl extension support

joshsleeper:
## changes

- add `hcl` extension support
    - this is used by most Hashicorp tools, including `consul`, `packer`, and `vault` to name a few
    - <https://github.com/hashicorp/hcl>
- alphabetize `EXTENSIONS` keys
    - they were mostly there already, but a few weren't quite right

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -62,6 +62,7 @@ EXTENSIONS = {
     'gypi': {'text', 'gyp', 'python'},
     'gz': {'binary', 'gzip'},
     'h': {'text', 'header', 'c', 'c++'},
+    'hcl': {'text', 'hcl'},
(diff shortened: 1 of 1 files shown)


asottile:
yeah some of the extensions in `EXTENSIONS` are intentionally not alphabetized so they can be grouped near their relevant things (for example the nim sources are near each other and csproj was next to csharp)

asottile:
one thought is to whether the existing terraform extensions should also set `hcl` -- but that can always be a followup

joshsleeper:
ah, sorry for the sorting noise then haha, thanks for the merge and fixing that for me!
I'd def thought about adding `hcl` to `tf` and `tfvars`, but figured I'd keep this simple and focused in case that was up for debate 😆

PR #234, opened 2021-09-28
Add DDL file extension as Text/SQL

1337Lurker:
### Changes
---
- [x] Add `ddl` extension support as Text/SQL file extension.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -35,6 +35,7 @@ EXTENSIONS = {
     'cuh': {'text', 'cuda'},
     'cxx': {'text', 'c++'},
     'dart': {'text', 'dart'},
+    'ddl': {'text', 'sql'},
     'def': {'text', 'def'},
     'dll': {'binary'},
     'dtd': {'text', 'dtd'},


asottile:
do you have a reference for this?  I scrolled through 10 pages of github searches for `*.ddl` and didn't find any ddl files which were sql (one was a proprietary format, the other was vbscript)

1337Lurker:
It's one of Jetbrains' DataGrip default file extensions, pictured [here](https://www.jetbrains.com/help/datagrip/creating-and-registering-file-types.html#change-extension) (second image down):

From my local IDE:

image

From my own experience, `sql` and `ddl` are used, but I'm not familiar with `cql`, and, as such` did not add it.

asottile:
unless you can come up with an official documented resource and some examples I can't accept this

1337Lurker:
Upon further review, it looks like `ddl` is only a local convention, and the larger convention is to only use `sql` and not differentiate between DCL/DDL/DML/DQL using file extensions.

Thanks for all the quick responses, @asottile, I am much obliged, sir.

PR #235, opened 2021-09-29
Handle shebangs with multiple arguments via env -S

### LINKED ISSUE #148: handle `env -S`
`env -S` is a way to pass more arguments in shebangs -- identify already shlex-splits the shebang so this should be as easy as ignoring `/usr/bin/env -S` at the beginning

example:

```console
$ cat t
#!/usr/bin/env -S python3 -u
print('hi')
$ ./t
hi
$ ~/opt/venv/bin/identify-cli t  # currently broken
["executable", "file", "text"]
```

it should produce:

```console
$ ~/opt/venv/bin/identify-cli t
["executable", "file", "python", "python3", "text"]
```

errsyn:
Resolves #148

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -193,7 +193,11 @@ def parse_shebang(bytesio: IO[bytes]) -> Tuple[str, ...]:
 
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
-        cmd = cmd[1:] if cmd[1] != '-S' else cmd[2:]
+        if cmd[1] == '-S':
+            cmd = cmd[2:]
+        else:
+            cmd = cmd[1:]
+
         if cmd == ('nix-shell',):
             return _parse_nix_shebang(bytesio, cmd)
     return cmd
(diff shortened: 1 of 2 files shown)


identify/identify.py
```diff
@@ -193,7 +193,7 @@ def parse_shebang(bytesio: IO[bytes]) -> Tuple[str, ...]:
 
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
-        cmd = cmd[1:]
+        cmd = cmd[1:] if cmd[1] != '-S' else cmd[2:]
```
asottile, line 196: I think the problem is slightly more subtle than this (and the current splitting may be incorrect)

I would also prefer these to be actual if statements as ternary hides the (potentially) missing coverage and is more difficult for a human to read quickly

errsyn: I could think of three instances, in which the proposed splitting logic would indeed fail us.

1. Multiple options to `env` and `-S` isn't the first one:
```
#!/usr/bin/env -v -S python -u
```

2. Combined options:
```
#!/usr/bin/env -Sv python -u
```

3. Using the long option instead of the short option:
```
#!/usr/bin/env --split-string=python -u
```

If we want to correctly handle the first and second example, we, more or less, would have to expand the scope of the issue to handle all options of `env`. The third example would further add handling their respective long option forms.

Were you thinking of other or more general cases, in which the current splitting might be insufficient or incorrect? If so, could you elaborate?

asottile (approved):
this is good enough for now -- if the more complex case comes up later then we can handle that separately

PR #239, opened 2021-10-20
Add generic binary file extension

Paluche:
Add the .bin extension support, for generic binary files.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -9,6 +9,7 @@ EXTENSIONS = {
     'bat': {'text', 'batch'},
     'bats': {'text', 'shell', 'bash', 'bats'},
     'bib': {'text', 'bib'},
+    'bin': {'binary'},
     'bmp': {'binary', 'image', 'bitmap'},
     'bz2': {'binary', 'bzip2'},
     'bzl': {'text', 'bazel'},


asottile:
I don't think this adds anything -- identify would already select such files as binary

Paluche:
Indeed, my issue is somewhere else.

Paluche:
@asottile I actually happen to have a file with the prefix .bin which is not identified as binary but text. I cannot share the content of the file. I'll investigate a bit further.

Paluche:
So the issue with my file, is that it is representing a binary flash content, with starts with several kB of `0xff` bytes, then at some point we have actual binary content. Due to identify saying the file is a not a binary, `end-of-file-fixer` is messing  with my file content.

Experimenting around with `file` command line (on Ubuntu), we apparently cannot expect to determinate a file with only `0xff` bytes inside is a binary file. 

We could enhance the `is_text()` method to obtain a better guessing. A simple solution would be to parse  the whole file or the first MB instead of the first kB. 1MB is from tests not a visible overload. Project linked in comments as inspiration for the `is_text()` method is typically running its algorithm on the whole file, see [here](https://github.com/file/file/blob/c07e242e766242a44ff720c149b1bdd4924ec247/src/encoding.c#L80).

Anyway  the change I was proposing is not indeed not the proper solution, nor can we obtain an deterministic answer for that type of file. I'll close that PR for good.

PR #252, opened 2021-11-17
go.mod and go.sum are now detected as golang

fkautz:
go.mod and go.sum are used to define module dependencies and checksums in golang.

adding these two files to the list of known extensions/files.

﻿Signed-off-by: Frederick F. Kautz IV <fkautz@alumni.cmu.edu>

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -288,6 +288,8 @@ NAMES = {
     'Gemfile': EXTENSIONS['rb'],
     'Gemfile.lock': {'text'},
     'GNUmakefile': EXTENSIONS['mk'],
+    'go.mod': {'text', 'go-mod'},
+    'go.sum': {'text', 'go-sum'},
     'Jenkinsfile': EXTENSIONS['jenkins'],
     'LICENSE': EXTENSIONS['txt'],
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -288,6 +288,8 @@
     'Gemfile': EXTENSIONS['rb'],
     'Gemfile.lock': {'text'},
     'GNUmakefile': EXTENSIONS['mk'],
+    'go.mod': EXTENSIONS['go'],
+    'go.sum': EXTENSIONS['go'],
```
asottile, line 292: I think these probably would get classified a little differently since they aren't actually "go" go source, but a different related format?

fkautz: That is correct, they are not go code. Do you have a recommendation for this? Perhaps we can call these something unique such as go-mod and go-sum?

asottile: yeah `{'text', 'go-mod'}` and `{'text', 'go-sum}` probably makes sense -- that way if a tool specifically targets those it can differentiate

fkautz: should be good to go now.

PR #261, opened 2021-12-28
Add .prettierignore extension

andreoliwa:
From https://prettier.io/docs/en/ignore.html#ignoring-files-prettierignore:

> .prettierignore uses gitignore syntax.


identify/extensions.py
```diff
@@ -269,6 +269,7 @@
     '.mention-bot': EXTENSIONS['json'] | {'mention-bot'},
     '.npmignore': {'text', 'npmignore'},
     '.pdbrc': EXTENSIONS['py'] | {'pdbrc'},
+    '.prettierignore': {'text', 'gitignore'},
```
asottile, line 272: this should also mention `prettierignore` specifically in case one wants to target that as well

asottile (approved):

PR #264, opened 2022-01-04
Add mjs as JavaScript file extension

adamchainz:
`.mjs` is the extension recommended for JavaScript modules by the V8 docs: https://v8.dev/features/modules#mjs , and the extension is used by other tools like Babel.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -113,6 +113,7 @@ EXTENSIONS = {
     'md': {'text', 'markdown'},
     'mdx': {'text', 'mdx'},
     'mib': {'text', 'mib'},
+    'mjs': {'text', 'javascript'},
     'mk': {'text', 'makefile'},
     'ml': {'text', 'ocaml'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -113,6 +113,7 @@
     'md': {'text', 'markdown'},
     'mdx': {'text', 'mdx'},
     'mib': {'text', 'mib'},
+    'mjs': {'text', 'javascript'},
```
asottile, line 116: should we additionally add `mjs` as a tag in case someone wants to target that specifically?

adamchainz: Maybe wait for a use case? I found MDN actually argues against using the `.mjs` suffix since many tools don't support it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js . So it may not be particularly popular.

asottile (approved):

PR #271, opened 2022-01-21
Adding `sty` files (normally TeX files)

tdegeus:
I was doubting to add a secondary entry `tex`, but I don't think this is justified. Some searching gave me also some non-TeX usage. 

Reference for usage with TeX: https://www.overleaf.com/learn/latex/Understanding_packages_and_class_files


asottile:
please provide references

tdegeus:
@asottile see edit

asottile:
visual basic also uses `.cls` files it appears so we probably can't add that one

tdegeus:
@asottile Better like this?

asottile (approved):

PR #277, opened 2022-01-31
Add `.hbs` extension for Handlebars templates

thibaudcolas:
[Handlebars](https://handlebarsjs.com/) templates most commonly use the `.hbs` extension. I believe [Ember](https://guides.emberjs.com/release/components/) is the most popular tool that uses `.hbs`. [Eleventy](https://www.11ty.dev/docs/languages/handlebars/) is another example.


asottile (approved):
it looks like handlebars also uses `.handlebars` -- at least from this:  https://github.com/handlebars-lang/handlebars.js/search?l=handlebars

PR #280, opened 2022-02-05
add poetry.lock as toml

kreyoo:
I noticed that the lock files from poetry and yarn aren't in here yet. Poetry uses toml. Yarn is another story. Yarn v1 files aren't valid yaml files while Yarn v2 lock files are. I think it's better to go just with a text file here.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -312,6 +312,7 @@ NAMES = {
     'Pipfile': EXTENSIONS['toml'],
     'Pipfile.lock': EXTENSIONS['json'],
     'PKGBUILD': {'text', 'bash', 'pkgbuild', 'alpm'},
+    'poetry.lock': EXTENSIONS['toml'],
     'pylintrc': EXTENSIONS['ini'] | {'pylintrc'},
     'README': EXTENSIONS['txt'],
     'Rakefile': EXTENSIONS['rb'],


identify/extensions.py
```diff
@@ -312,10 +312,12 @@
...
     'Pipfile.lock': EXTENSIONS['json'],
     'PKGBUILD': {'text', 'bash', 'pkgbuild', 'alpm'},
+    'poetry.lock': EXTENSIONS['toml'],
     'pylintrc': EXTENSIONS['ini'] | {'pylintrc'},
     'README': EXTENSIONS['txt'],
     'Rakefile': EXTENSIONS['rb'],
     'setup.cfg': EXTENSIONS['ini'],
     'WORKSPACE': EXTENSIONS['bzl'],
     'wscript': EXTENSIONS['py'],
+    'yarn.lock': {'text'},
```
asottile, line 322: this doesn't improve the identification -- it will get `text` already

kreyoo: I oriented this change on the Gemfile.lock entry:
https://github.com/pre-commit/identify/blob/939e9c9dab1217a54eedd6ac6e95273a0fbdfd66/identify/extensions.py#L300
What exactly is the difference between Gemfile.lock and yarn.lock?

Furthermore, checking yarn.lock with the cli returns no result:
```
identify /home/janv/yarn.lock 
identify: no decode delegate for this image format `LOCK' @ error/constitute.c/ReadImage/572.
```

asottile: your output is doctored, you're running imagemagick identify

the `Gemfile.lock` is a mistake, it seems -- it should have a more specific tag

kreyoo: Oh... my bad. I had trouble getting the cli to run  and got something wrong somehwhere. How about adding a `__main__.py` so the cli can be run with `python -m identify` ? I will revert the `yarn.lock` change.

asottile: I mean you've kinda killed my trust -- why did you make up output?

the command is called `identify-cli` to disambiguate with imagemagick

kreyoo: I did the stupid mistake to not read the docs/README. I know this isn't trustworthy and I want to apologize for that. I was so sure that `identify` was the right cli command so I stripped off the `python -m` before it. I didn't pay enough attention to what I am doing. Lesson learned. 

However, I want to regain your trust by pointing out the code where `poetry.lock` is interpreted as a `toml` file:
https://github.com/python-poetry/poetry/blob/31657e7e5c47da758c14645035e59eb7813ac2f8/src/poetry/packages/locker.py#L53

Again, sorry for killing your trust. This was embarrassing.

asottile (approved):

PR #282, opened 2022-02-07
identify: check shebang of all text files

rc-matthew-l-weber:
There are cases where shell scripts (and other scripts) may not be executable but still need to be checked.

* Remove early os.access() check which will return (never get to shebang check) for shell scripts that are not executable
* Check that a file is text first and then attempt to parse the shebang

Signed-off-by: Matthew Weber <matthew.weber@collins.com>

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -65,7 +65,7 @@ def tags_from_path(path: str) -> set[str]:
     if len(t) > 0:
         tags.update(t)
     else:
-        if executable:
+        if file_is_text(path):
             shebang = parse_shebang_from_file(path)
             if len(shebang) > 0:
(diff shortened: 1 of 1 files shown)


asottile:
in the future please search existing issues and start with an issue

rc-matthew-l-weber:
I didn't see an existing issue that matched this.  Are you suggesting I open a ticket and describe the issue/reference my patch?

asottile:
yes, but don't bother because this patch will not be accepted -- shebangs only have an effect if a file is executable and the implementation is *intentional* to that -- it should have been fairly obvious when you had a failing test demonstrating exactly that behaviour

PR #285, opened 2022-02-22
support identifying qml files as qml language

jamesharris-garmin:
The Qt project provides a language called qml that is provided in .qml files. This language is common enough that it should be identifiable via tags in order to allow pre-commit and other projects to write hooks for those files.


asottile:
can you provide a reference for the extension showing its unambigious use -- thanks!

jamesharris-garmin:
https://doc.qt.io/qt-5/qtqml-documents-topic.html

asottile (approved):

jamesharris-garmin:
When can we release this to pypi?

asottile:
35 minutes **before** your comment, lol

PR #290, opened 2022-03-16
(Slightly) expand executable check to include os.stat

posita:
Use of virtualized platforms is increasing in popularity. This includes using them to
create consistent environments for local development, despite underlying platforms
potentially varying widely. Even within a single organization of non-trivial size, it's
not unusual to have developers running several different versions of a single OS on
several different hardware revs from the same manufacturer. Containerization is a
popular way to attempt to eliminate environmental differences to achieve consistency.

Some virtualization layers confuse permissions checks inside virtualized environments
when those environments have access to host filesystems. docker/for-mac#5509 is a prime
example.

This expands the ways in which file executability is detected slightly. The prior
os.access check remains. If that should fail, this change includes an os.stat fallback.
Together, these should get closer to detecting whether the file is truly executable on
systems where such confusion is present.

Fixes 289.

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -53,7 +53,12 @@ def tags_from_path(path: str) -> set[str]:
 
     tags = {FILE}
 
-    executable = os.access(path, os.X_OK)
+    # Some systems' access(2) don't correctly surface whether a file is
+    # executable, so we fall back to another check to be sure.
+    executable = (
+        os.access(path, os.X_OK) or
+        bool(os.stat(path).st_mode & stat.S_IXUSR)
+    )
     if executable:
         tags.add(EXECUTABLE)
     else:


posita:
@asottile, I understand you don't want to accept this, but can you offer some insight as to why?

posita:
I did search. I thought I [made a case](https://github.com/pre-commit/identify/issues/289) for why [your response to #189](https://github.com/pre-commit/identify/pull/189#issuecomment-808854232) should be revisited, and offered, what I thought were novel arguments in favor of accommodating your users with minimal intrusion into code integrity. I accept those may not have been persuasive, but I have no insight into your reasoning. You closed the issue and this PR so quickly, that I'm not even sure they were considered?

asottile:
I hope you understand that questioning past decisions for revisiting without providing new arguments is a waste of everyones time.  please do not do that

posita:
I'm a bit confused about this. From my perspective, my arguments _were_ novel, as was my proposed solution. If you look at my proposed diff and compare it to that which was rejected in #189, you'll see it's different. I attempted to walk a line between not offering a solution to users (your default position) and something less of a deviation in #189 that preserved original behavior. I think you closed each of this PR and the corresponding issue within like 60 seconds of their respective filings. Maybe they weren't novel after all, or maybe I didn't make the differences or nuances I presented accessible, but providing something new was my understanding of what I was doing.

asottile:
the decision in the duplicate is "we're not hacking around a mistake made by docker inc"

PR #299, opened 2022-04-26
Add fish shell interpreter

benspaulding:
This adds [fish](https://fishshell.com) to the interpreters.

In a separate commit I also marked the `fish` entries as `shell` in case omitting it was an oversight. If that was a deliberate decision we can drop the commit.

Let me know if you have any concerns or need anything else from me to move this forward!

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -51,7 +51,7 @@ EXTENSIONS = {
     'exe': {'binary'},
     'eyaml': {'text', 'yaml'},
     'feature': {'text', 'gherkin'},
-    'fish': {'text', 'fish'},
+    'fish': {'text', 'shell', 'fish'},
     'gd': {'text', 'gdscript'},
     'gemspec': {'text', 'ruby'},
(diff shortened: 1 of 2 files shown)


benspaulding:
> I'm not sure about either of the commits: 
> * fish isn't intended as a scripting language like bash is

I didn't see any kind of documentation on what did or did not belong in the interpreters, so I submitted this PR in hopes that pre-commit could run on my fish scripts. For what it's worth, fish [documents itself as a scripting language](https://fishshell.com/docs/current/index.html) and that is the main way I and other developers I know use it.

> * `shell` is used by shellcheck for which `fish`'s syntax doesn't make sense

Cool, that makes sense. If you think the other commit adding fish as an interpreter is beneficial feel free to just take that commit. If you would like me to handle it I can push a commit reverting the shell one, or push a new branch with just that one commit, or force-push to this branch to remove the shell commit -- whatever works for you. :)

And I forgot to say earlier that I found the identify codebase very approachable and easy to contribute to — thanks for making it that way!

PR #300, opened 2022-04-26
Just interpreter, extension, and names

benspaulding:
This adds the `justfile` file type for the [`just`](https://github.com/casey/just) program.

A second commit also add `just` as an interpreter as that is a [documented](https://github.com/casey/just#just-scripts) and reasonably common use of `just`.


identify/extensions.py
```diff
@@ -304,6 +306,8 @@
     'go.mod': {'text', 'go-mod'},
     'go.sum': {'text', 'go-sum'},
     'Jenkinsfile': EXTENSIONS['jenkins'],
+    'Justfile': EXTENSIONS['just'],
```
asottile, line 309: `just` appears to be case insensitive -- I don't think identify should support `Justfile` given the documented examples are all `justfile` or `.justfile`

asottile (commented):
why 2 commits?

benspaulding:
> why 2 commits?

In my mind it made sense to break them up in case one part or the other caused problems or was not wanted; no reason beyond that. If you decide to merge this PR as-is squashing should be fine right? Or is there a particular reason to avoid two commits in a PR in the identify repo?

identify/extensions.py
benspaulding: > `just` appears to be case insensitive -- I don't think identify should support `Justfile` given the documented examples are all `justfile` or `.justfile`

There are other filenames that are case-insensitive, such as README/ReadMe/Readme, LICENSE — I should have noticed that identify only handles one case for them and followed that pattern.

Would you like me to remove that line in another commit, or is that something you would like to handle? (I ask because it seems you don't like multiple commits.)

The other option may be for a bigger change of being able to mark particular names as being case-insensitive. (But not in this PR, obviously.)

PR #301, opened 2022-04-26
V (a.k.a. vlang) interpreter and extensions

benspaulding:
This adds the `v` file type, as well as the alias `vlang`, to the `.v` extension for the [V Programming Language](https://vlang.io).

A second commit also adds the `.vsh` extension and the `v` interpreter for [V scripts](https://github.com/vlang/v/blob/master/doc/docs.md#cross-platform-shell-scripts-in-v).

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -211,7 +211,7 @@ EXTENSIONS = {
     'txsprofile': {'text', 'ini', 'txsprofile'},
     'txt': {'text', 'plain-text'},
     'urdf': {'text', 'xml', 'urdf'},
-    'v': {'text', 'verilog'},
+    'v': {'text', 'verilog', 'v', 'vlang'},
     'vb': {'text', 'vb'},
     'vbproj': {'text', 'xml', 'vbproj'},
     'vcxproj': {'text', 'xml', 'vcxproj'},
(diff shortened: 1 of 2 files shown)


identify/extensions.py
```diff
@@ -211,7 +211,7 @@
     'txsprofile': {'text', 'ini', 'txsprofile'},
     'txt': {'text', 'plain-text'},
     'urdf': {'text', 'xml', 'urdf'},
-    'v': {'text', 'verilog'},
+    'v': {'text', 'verilog', 'v', 'vlang'},
```
asottile, line 214: ah yeah we can't have an extension mapping to two different tags

benspaulding: > ah yeah we can't have an extension mapping to two different tags

I saw others like `svg` that are both image and XML and thought that was okay, but I don't have a full understanding of identify, so thanks for calling it out!

```py
'svg': {'text', 'image', 'svg', 'xml'},
```

My other thought was that it is unlikely for a project to be working with _both_ Verilog _and_ V files because that would be a headache, but this must not work for other reasons as well? 🤷🏻‍♂️

Does this mean that there is no way for identify to support V? Do you have ideas of how we could handle this?

Thanks for your help!

PR #308, opened 2022-06-08
Implement Snakemake extensions

sappjw:
Snakemake is a workflow management system with workflows described via a
human-readable, Python-based language. See
https://snakemake.readthedocs.io.


identify/extensions.py
```diff
@@ -182,6 +182,7 @@
     'sh': {'text', 'shell'},
     'sln': {'text', 'sln'},
     'sls': {'text', 'salt'},
+    'smk': {'text', 'snakefile'},
```
asottile, line 185: googling "smk extension" doesn't lead to snakefile unfortunately -- so I'm not sure we can accept this :(

PR #314, opened 2022-07-15
Add FITS files to extension list

mcoughlin:
This PR adds FITS (https://fits.gsfc.nasa.gov/) files to the extension list.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -52,6 +52,7 @@ EXTENSIONS = {
     'eyaml': {'text', 'yaml'},
     'feature': {'text', 'gherkin'},
     'fish': {'text', 'fish'},
+    'fits': {'binary', 'fits'},
     'gd': {'text', 'gdscript'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -52,6 +52,8 @@
     'eyaml': {'text', 'yaml'},
     'feature': {'text', 'gherkin'},
     'fish': {'text', 'fish'},
+    'fit': {'binary', 'fits'},
```
asottile, line 55: searching for this file extension produces some ambiguity so I don't think this particular one can be accepted

mcoughlin: So you want fits but not fit or what's the plan?

asottile: yeah fits seems unambiguous

mcoughlin: Ok removed.

asottile (approved):

Issue #100, opened 2020-03-26
feature request: add `.env` files


ghostsquad:
This is a feature request to add `.env` files to identify

asottile:
what should `.env` be classified as?

ghostsquad:
`env` maybe? or `dotenv` ?

Here's some popular libraries/tools that use files like this:

https://github.com/motdotla/dotenv?ts=4

https://github.com/joho/godotenv?ts=4

https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

https://parceljs.org/env.html

asottile:
sure, would you like to add it?  it's a little odd since usually you don't check in the `.env` file

I think this would go [here](https://github.com/chriskuehl/identify/blob/b8a3ad33b3b6e26da11b03d5f577f920710c2d18/identify/extensions.py#L184)?

ghostsquad:
>  it's a little odd since usually you don't check in the .env file

that's actually not true. It's quite convenient to check in .env files that contain non-secret information for development & testing environments. In fact, most of the tools above support varying features for overriding values. Like in `godotenv` you apply .env files in order of most specific to least specific like:

```
local.env
development.env
global.env
```

and once a value is set, it is _not_ overwritten (which is why you do it in this order). But other tools do the opposite (so definitely read the docs! 📖 )

But anyways, if your application follows the 12-factor methodology: https://12factor.net/

Then likely it's simply easiest to define your config in this way.

ghostsquad:
> I think this would go [here](https://github.com/chriskuehl/identify/blob/b8a3ad33b3b6e26da11b03d5f577f920710c2d18/identify/extensions.py#L184)?

Yes, likely. Any file ending in `.env` though, there's a small concern that existing projects may not use this convention, as I've seen `.env.test` and similar before.

also, apparently, I haven't entirely even read all the documentation here:
https://github.com/motdotla/dotenv?ts=4#should-i-have-multiple-env-files

Which suggests not use multiple, overriding env files. I'm not entirely sure I agree, but that's ok.

asottile:
- https://github.com/motdotla/dotenv?ts=4#should-i-commit-my-env-file
- https://github.com/motdotla/dotenv?ts=4#should-i-have-multiple-env-files

both seem to suggest you should (1) not check it in and (2) not have cascading

ghostsquad:
The only reason is says not to check them in, is because they may contain secrets.

[...]

PR #317, opened 2022-08-02
Added env/dotenv file extension

Cediddi:
Adding env file extension means, env file validators can run on pre-commit. While it's a bad practice to send .env files to git repo, examples or encrypted templates are very useful.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -45,6 +45,7 @@ EXTENSIONS = {
     'ear': {'binary', 'zip', 'jar'},
     'edn': {'text', 'clojure', 'edn'},
     'ejs': {'text', 'ejs'},
+    'env': {'text', 'dotenv'},
     'eot': {'binary', 'eot'},
     'eps': {'binary', 'eps'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -45,6 +45,7 @@
     'ear': {'binary', 'zip', 'jar'},
     'edn': {'text', 'clojure', 'edn'},
     'ejs': {'text', 'ejs'},
+    'env': {'text', 'env', 'dotenv'},
```
asottile, line 48: should be one or the other, not both

asottile:
also just making sure you read the discussion about this in #100

identify/extensions.py
Cediddi: Alright, removed env and kept dotenv (and replaced the commit)

Cediddi:
Indeed, I read the related issue, I even commented before opening this MR. I'm writing my second comment there right now :)

asottile (approved):

PR #324, opened 2022-09-05
Identify `.importlinter` as an INI-file

webknjaz:
_(documented at https://import-linter.readthedocs.io/en/stable/usage.html#configuration-file-location)_

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -280,6 +280,7 @@ NAMES = {
     '.gitlint': EXTENSIONS['ini'] | {'gitlint'},
     '.gitmodules': {'text', 'gitmodules'},
     '.hgrc': EXTENSIONS['ini'] | {'hgrc'},
+    '.importlinter': EXTENSIONS['ini'] | {'import-linter'},
     '.isort.cfg': EXTENSIONS['ini'] | {'isort'},
     '.jshintrc': EXTENSIONS['json'] | {'jshintrc'},
     '.mailmap': {'text', 'mailmap'},


asottile:
this doesn't seem popular enough or specific enough to make all classifications of "importlinter" ini

webknjaz:
Well, the nitpick tool depends on identify to see if a file is ini. They don't provide any mechanisms for defining whether a given file is INI. Besides, identify already lists files that aren't popular so such a judgment seems weird at best. Identify doesn't have a mechanism for extending the list externally, does it?

asottile:
everything so far seems pretty popular -- I think my core issue is that `importlinter` is an extremely generic name and is only ~200 stars so it should probably use a more specific name rather than "importlinter".

webknjaz:
> everything so far seems pretty popular

Here's an immediate example — https://github.com/pre-commit/identify/blob/207d5f7/identify/extensions.py#L286. It's a config for a bot that Facebook used to run. It was useful half a decade ago but has been sunset for years already.

> so it should probably use a more specific name rather than "importlinter".

There are other config fallbacks but I prefer this one because mixing up multiple tool configs in the same file is a can of worms. FWIW I don't think that any other tool is using the same filename so it'd be pretty safe to identify it for what it is.

asottile:
> > everything so far seems pretty popular
> 
> Here's an immediate example — https://github.com/pre-commit/identify/blob/207d5f7/identify/extensions.py#L286. It's a config for a bot that Facebook used to run. It was useful half a decade ago but has been sunset for years already.

... which surprise surprise was [added half a decade ago](https://github.com/pre-commit/identify/commit/bf87804e4f7c1d95e5cea2023be2217dd2e15bea)

PR #330, opened 2022-09-29
Add .patch file extension in a patch type

gnought:


asottile:
what's the goal here? does `types: [diff]` not do what you want?

gnought:
Yes, you're right. I misread the value is a list of file extensions.

Issue #334, opened 2022-10-04
Identify jsonld and yamlld


ioggstream:
## I expect

To identify the following Linked Data extensions

- yamlld -> yaml https://json-ld.github.io/yaml-ld/spec/#application-ld-yaml
- jsonld -> json https://www.iana.org/assignments/media-types/application/ld+json

PR #335, opened 2022-10-04
Fix: #334. Add .yamlld and .jsonld extensions.

ioggstream:
## This PR

- adds .yamlld and .jsonld extensions


asottile:
these should probably be classified as both their `ld` equivalents and the base config language as well in case one would want to lint those specifically

ioggstream:
Hi @asottile! I landed here from pre-commit.

1. The desiderata is that jsonld and yamlld are at least linted with the JSON/yaml linters. It seems that check-yaml skips .yamlld files 

2. It is possible to add further validation (eh using jsonld) JSON schema but it is a complex task to be achieved consistently.

How do you think it is possible to achieve point 1?

Have a nice day!

asottile:
your patch does part 1 but you need to address my comment to get merged

ioggstream:
PTAL I "copied" from the xhtml entry. Have a nice day, R.

asottile (approved):

PR #338, opened 2022-10-20
feat(binfiles): add bin, elf, and obj files

adam-moss:
Signed-off-by: Adam Moss <178271-adam-moss@users.noreply.gitlab.com>

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 EXTENSIONS = {
     'adoc': {'text', 'asciidoc'},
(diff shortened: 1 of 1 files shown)


asottile (commented):
please provide references to your extensions -- they need to unambiguously refer to something

I don't think bin and obj qualify for this for example -- and I've not seen anything with `.elf` before

asottile:
yeah like I said, bin and obj are out because they are ambiguous -- and I've not seen `.elf` used -- only extensionless elf binaries

asottile:
also search the tracker next time please -- #239 

and don't use conventional commits, they're pure noise

PR #340, opened 2022-10-25
Erlang extensions and files

sebastiw:
Most common extensions for [Erlang](https://www.erlang.org/docs/25/man/erlc.html#supported-compilers)
I also included a configuration file `rebar.config` for one of the most common compilers. The other files included are `sys.config` specifying the released applications configuration and `sys.config.src` for the pre-released version.


identify/extensions.py
```diff
@@ -10,6 +10,7 @@
     'bash': {'text', 'shell', 'bash'},
     'bat': {'text', 'batch'},
     'bats': {'text', 'shell', 'bash', 'bats'},
+    'beam': {'binary', 'erlang'},
```
asottile, line 13: I don't think this is quite right -- if one were to try and filter erlang source files and they got a `.beam` file I think the linter would be quite surprised?

identify/extensions.py
```diff
@@ -78,6 +80,7 @@
     'hcl': {'text', 'hcl'},
     'hh': {'text', 'header', 'c++'},
     'hpp': {'text', 'header', 'c++'},
+    'hrl': {'text', 'header', 'erlang'},
```
asottile, line 83: while the current `header` tag isn't great -- I think it kinda means C-like headers so this would break tools expecting those?

identify/extensions.py
sebastiw: Ah sorry yeah you are right, I'll update.

identify/extensions.py
sebastiw: I'm not 100% sure what you mean. Erlang header files usually contain macros shared between source files. Do you have an example on how it's used in c?

asottile: it's that it's already used for c likes so reusing the tag for erlang would break existing things

sebastiw: hm, ok. I've removed 'header' for now, let's see how it works and in worst case I'll add a 'header_erlang' tag or something then.

PR #341, opened 2022-10-27
Identify obj and mtl files

MatthewRFennell:
This commit adds support for [Wavefront](https://en.wikipedia.org/wiki/Wavefront_.obj_file) and the associated [Material Template Library](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Material_template_library) files.


asottile:
obj has already been rejected in the past since it is far too generic

MatthewRFennell:
Ok, my bad. Thanks for taking a look.

PR #352, opened 2022-12-14
Add ppm file to extensions

andrewring:
PPM is portable pixmap format

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -267,6 +267,7 @@ EXTENSIONS = {
 }
 EXTENSIONS_NEED_BINARY_CHECK = {
     'plist': {'plist'},
+    'ppm': {'image', 'ppm'},
 }
 
 NAMES = {


identify/extensions.py
```diff
@@ -163,6 +163,7 @@
     'png': {'binary', 'image', 'png'},
     'po': {'text', 'pofile'},
     'pp': {'text', 'puppet'},
+    'ppm': {'binary', 'image', 'ppm'},
```
asottile, line 166: the point of ppm is it's _not_ binary https://anthonysottile.com/ppm/

andrewring: As I look more, [it's actually conditional](https://en.wikipedia.org/wiki/Netpbm). The magic number at the beginning of the file indicates if it should be parsed as ASCII or binary, so I guess we can't really make this kind of assertion based on extension alone, unfortunately. Is there somewhere in the existing flow that this type of check would be supported? The linux `file` utility, for example, can correctly identify a binary example, like this:
```
frame-40.ppm: Netpbm image data, size = 1920 x 1080, rawbits, pixmap
```

asottile: https://github.com/pre-commit/identify/blob/eb5467839d10005bbdf59c21e989c07013dcc750/identify/extensions.py#L268 it is then

andrewring: Updated. Please let me know if there's somewhere else that also needs updating.

asottile (approved):

PR #361, opened 2023-01-11
Adds file extensions for RDF and SPARQL

Tpt:
The extensions are taken from the relevant W3C specifications and the human-readable names are derived from the MIME type "subtype" part

* nq: https://www.w3.org/TR/2014/REC-n-quads-20140225/
* nt: https://www.w3.org/TR/2014/REC-n-triples-20140225/
* rdf: https://www.w3.org/TR/2014/REC-rdf-syntax-grammar-20140225/
* rq: https://www.w3.org/TR/2013/REC-sparql11-query-20130321/
* ru: https://www.w3.org/TR/2013/REC-sparql11-update-20130321/
* srj: https://www.w3.org/TR/2013/REC-sparql11-results-json-20130321/
* srx: https://www.w3.org/TR/2013/REC-sparql11-results-json-20130321/
* trig: https://www.w3.org/TR/2014/REC-trig-20140225/
* ttl: https://www.w3.org/TR/2014/REC-turtle-20140225/


Tpt:
All these files types are registered at IANA so should hopefully be fairly stable. But indeed, there seems to be old formats using some of them.

Would keeping only "rdf", "trig", "rq" and "ttl" fine for you? "rdf", "trig" and "ttl" seems fairly unambiguus and "rq" is aleady bound to SPARQL by [pygments](https://github.com/pygments/pygments/blob/fd0c3e9e0bc6a2a0101b0b16e78bb9594e1cf9fd/pygments/lexers/_mapping.py#L463).

identify/extensions.py
```diff
@@ -183,8 +183,10 @@
     'r': {'text', 'r'},
     'rake': {'text', 'ruby'},
     'rb': {'text', 'ruby'},
+    'rdf': {'text', 'xml', 'rdf-xml'},
```
asottile, line 186: idk that the `-xml` adds anything here -- unless there's a reason to have a distinction many of the tags just match their extensions

Tpt: Thank you for you review! It's to follow the common naming of the format: https://en.wikipedia.org/wiki/RDF/XML

asottile: is there a non-xml rdf?

Tpt: Yes, [Turtle](https://www.w3.org/TR/2014/REC-turtle-20140225/) (`.ttl`, a plain text format), [N-Triples](https://www.w3.org/TR/2014/REC-n-triples-20140225/) (`.nt`, a CSV-like format) and [JSON-LD](https://www.w3.org/TR/json-ld/) (`.json-ld`, already in the list, a JSON-based format).

Tpt:
@asottile Sad to see this MR closed. Is there something I can do to add the RDF/SPARQL related file extension to identify. thank you again for this great library.

asottile:
my request went unanswered for 10 months

Tpt:
> my request went unanswered for 10 months

So sorry for that. I guess you mean this request: <q>can you update the PR to only include ones which are unambigious?</q>
If yes, I have updated the MR after your request to only cover  "rdf", "trig", "rq" and "ttl" (see my previous comment on the rational why). But maybe there is an other request I have missed. So sorry for that if it's the case.

Thank you again for all your amazing work and sorry for bother you.

PR #362, opened 2023-01-11
FreePascal and Lazarus extensions

ClasherKasten:
The [FreePascal Wiki](https://wiki.freepascal.org/File_extensions) describes some file extensions which could be worth adding to identify:
- `.pas`: Pascal source file (interchangeable with `.pp`, but that is already taken by puppet)
- `.lpi`: Lazarus project information, xml
- `.lpr`: Special pascal source file, marks main program
- `.lfm`: Lazarus form


identify/extensions.py
```diff
@@ -120,10 +120,13 @@
     'lean': {'text', 'lean'},
     'lektorproject': {'text', 'ini', 'lektorproject'},
     'less': {'text', 'less'},
+    'lfm': {'text', 'lazarus', 'form'},
```
asottile, line 123: should this be `lazarus-form` ?

ClasherKasten: I think we need to preserve the `lazarus` tag, as `.lfm` is a Lazarus specific file, but I changed `form` to `lazarus-form`.

asottile (approved):

PR #364, opened 2023-01-24
Add file type support for CAN database files (.dbc)

hofbi:
This adds `.dbc` files also known as CAN database files to the the supported file extensions. See https://docs.fileformat.com/database/dbc/ as a reference.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -41,6 +41,7 @@ EXTENSIONS = {
     'cxx': {'text', 'c++'},
     'cylc': {'text', 'cylc'},
     'dart': {'text', 'dart'},
+    'dbc': {'text', 'dbc'},
     'def': {'text', 'def'},
     'dll': {'binary'},
     'dtd': {'text', 'dtd'},


asottile:
I'm not sure this is unambigious enough -- from searching I see at least 4 different technologies which use this extension

hofbi:
But are all of them text based? If yes, does it then matter?

asottile:
are they? can you confirm that?

hofbi:
I can not, my search only finds CAN related stuff. Also found https://docs.openvehicles.com/en/latest/components/vehicle_dbc/docs/dbc-primer.html which indicates that my care manufacturers are using this.

What were the others you found?

asottile:
a "dbc extension" google search gives a bunch of stuff before CAN :S

hofbi:
That helped. I used "dbc file" and that resulted only in the CAN ones. With your suggestion I also found 

* https://fileinfo.com/extension/dbc with popularity low according to https://extension.nirsoft.net/dbc
* https://www.file-extension.org/extensions/dbc

With that search, I also found a VS Code extension syntax highlighting the CAN files: https://marketplace.visualstudio.com/items?itemName=lharri73.dbc

In general, I would say `.dbc` files are ambigious. But I also think that in particular for pre-commit users, the text-based CAN files are most likely. What do you think?

PR #370, opened 2023-02-24
Add .pdb file extension

noralutz:
Protein Data Bank (PDB) format is a standard for files containing atomic coordinates. https://www.wwpdb.org/documentation/file-format


asottile:
unfortunately, pdb is ambiguous (there's a much more popular binary format used by msvc) so we can't accept a definition for it

PR #376, opened 2023-03-16
Adds support for with-contenv of s6-overlay

stumpylog:
[s6-overlay](https://github.com/just-containers/s6-overlay) is a popular container process supervision suite.  It provides two shebang values for running a given command with the container's environment: `#!/usr/bin/with-contenv` or `#!/command/with-contenv`.  Both are essentially the same thing as `#!/usr/bin/env` but for containers.

This PR adds support for identifying the command run with these shebangs.  As the two styles/types exist, an `in` check instead of exactly equal was used (but could be updated)

### DIFF
--- a/identify/identify.py
+++ b/identify/identify.py
@@ -191,7 +191,7 @@ def parse_shebang(bytesio: IO[bytes]) -> tuple[str, ...]:
             return ()
 
     cmd = tuple(_shebang_split(first_line.strip()))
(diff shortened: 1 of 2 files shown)


stumpylog:
I could open one now, but with the PR already existing, that seems a little redundant at this point.

asottile:
s6 is not nearly as ubiquitous as `env` -- and `in` is not appropriate here

PR #378, opened 2023-03-19
Add justfiles support

loamhoof:
Adding support for justfiles, from the [just command runner](https://github.com/casey/just)

_About the filenames:_

`just` is case-insensitive when looking for a justfile. From the the [Quick Start section](https://just.systems/man/en/chapter_18.html):

> The search for a justfile is case insensitive, so any case, like Justfile, JUSTFILE, or JuStFiLe, will work.

Since there is no mechanism to add case-insensitive filenames, I added what I thought would be the most-common spellings: `Justfile` and `justfile` (the latter also adding support for "hidden" `.justfile` files).

_About the extension:_

While the tool doesn't bring specific support for the `.just` extension, it is the extension mentioned across the documentation and used by [all the examples](https://github.com/casey/just/blob/master/examples/) such as this [pre-commit.just](https://github.com/casey/just/blob/master/examples/pre-commit.just)

Additionally, a search for other tools using the `.just` extensions led to two things:

* [A ransomware](https://howtofix.guide/just-virus/): I believe it is safe to ignore as files encrypted by a ransomware should probably not be committed in any project
* [Some file from the Apple II series](https://www.solvusoft.com/en/file-extensions/file-extension-just/): I would say this one seems also pretty safe to ignore

Hence I believe there is no ambiguity with the `.just` extension


asottile:
please search for duplicates #300

PR #379, opened 2023-03-21
Identify config.ru rackup file as ruby

mmlb:


asottile:
can you provide a reference?  I'm having a bit of a difficult time finding what uses this

mmlb:
Hey @asottile here you go https://github.com/rack/rack#usage. I'm not super into the ruby web community but do work on a rack/rails app (I guess...?) a little bit for work. I'm not sure if it makes sense to limit to config.ru (GH search shows a few files that look like ruby that aren't explicitly named config.ru), GH's linguist just goes off of the extension .ru: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L5968

Would it be better to to just go with the extension?

asottile (approved):

mmlb:
:tada:

PR #381, opened 2023-04-19
Add php to interpreters list

signe:
Add php to interpreters list


asottile:
apparently I can't release this due to a pypi outage -- will try later

PR #389, opened 2023-06-08
Add extensions for Butane and Ignition configs

kytta:
Butane is a file format for human-readable OS configurations. Butane gets compiled to machine-readable Ignition config, which gets applied on the machine being provisioned.

[Butane files are YAML files](https://coreos.github.io/butane/config-fcos-v1_5/), and [Ignition files are JSON files](https://coreos.github.io/ignition/configuration-v3_4/).


asottile:
unfortunately neither of these extensions are unambigious -- googling either of them does not land on either of the technologies listed here

so unfortunately I can't accept this PR -- but thanks anyway!

kytta:
Yep, you're right. Honestly, I have no clue why they even do their own file extensions 🥴

PR #390, opened 2023-06-09
Add Visual Studio project file filters to extensions

renefritze:


identify/extensions.py
```diff
@@ -243,6 +243,7 @@
     'vb': {'text', 'vb'},
     'vbproj': {'text', 'xml', 'vbproj'},
     'vcxproj': {'text', 'xml', 'vcxproj'},
+    'vcxproj.filters': {'text', 'xml', 'vcxproj'},
```
asottile, line 246: this will not work as written

renefritze: Could you elaborate?

renefritze: I guess [this](https://github.com/pre-commit/identify/actions/runs/5222287345/jobs/9427575937?pr=390#step:7:47) is why.

identify/identify.py
```diff
@@ -85,7 +86,7 @@ def tags_from_path(path: str) -> set[str]:
 
 def tags_from_filename(path: str) -> set[str]:
     _, filename = os.path.split(path)
-    _, ext = os.path.splitext(filename)
+    ext = ''.join(Path(filename).suffixes)
```
asottile, line 89: this is not acceptable

asottile (changes requested):

identify/identify.py
renefritze: Can you please say why? Because it's using `pathlib`?

asottile: because it's a breaking change and in general incorrect

renefritze:
Let's take a step back. Can we agree that being able to assign tags for files like `foo.vcxproj.filters` would be useful?
And it's not practical to base assignment just on the `.filters` bit then, right? Which means some new functionality would need to enter the library. Can you help me do that in a way that is acceptable to you?

asottile:
identify does not intend to be a complete and perfect solution -- one stance it takes is that file extensions are a single part of the filename

renefritze:
> identify does not intend to be a complete and perfect solution -- one stance it takes is that file extensions are a single part of the filename

I feel like it would've been fairer towards me if that was your first comment instead of [this](https://github.com/pre-commit/identify/pull/390#discussion_r1224249671)

> this will not work as written

If you'd lead with that you did not want support files with "multi-part" suffixes I would not have spent time on trying to find an implementation that you found acceptable, make tests pass and add new ones.

asottile:
sure, but also I should not have needed to do that at all.  you could have _even tried your original PR_ and would have noticed it didn't work, would have noticed that the implementation does not support that, and stopped at that without creating the PR at all!

renefritze:
> sure, but also I should not have needed to do that at all. you could have _even tried your original PR_ and would have noticed it didn't work, would have noticed that the implementation does not support that, and stopped at that without creating the PR at all!

I don't think that logic parses. If I don't assume just adding a new file extension will work right away, but instead try this out locally, how would I conclude from that that adding changes to make it work is unacceptable? 

What do you think about (me) adding a note to the source file that states patterns like `.foo.bar` are intentionally not supported? To prevent someone trying the similar changes in the future?

PR #395, opened 2023-06-23
Add fmf extension

martinhoyer:
[.fmf](https://fmf.readthedocs.io/en/stable/concept.html#files) files are yaml text files used most notably in [tmt](https://github.com/teemtee/tmt).

Could for example be used in [tmt-lint pre-commit hooks](https://github.com/teemtee/tmt/blob/main/.pre-commit-hooks.yaml) or when running yaml linters against .fmf files.

See: https://github.com/teemtee/tmt/discussions/2164


asottile:
unfortunately searching this extension does not land on the tool you're suggesting so it is too ambiguous for inclusion here

PR #401, opened 2023-07-21
Add "json" type to jupter notebooks

connortann:
Suggest adding the "JSON" type to Jupyter notebooks  (`.ipynb` files), as notebooks are JSON files.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -99,7 +99,7 @@ EXTENSIONS = {
     'inl': {'text', 'inl', 'c++'},
     'ino': {'text', 'ino', 'c++'},
     'inx': {'text', 'xml', 'inx'},
-    'ipynb': {'text', 'jupyter'},
+    'ipynb': {'text', 'jupyter', 'json'},
     'j2': {'text', 'jinja'},
     'jade': {'text', 'jade'},
     'jar': {'binary', 'zip', 'jar'},


asottile (approved):

FelixSchwarz:
I just noticed that this PR did cause quite some fallout for us: Since `identify` now treats `ipynb` files as `json`, the `pre-commit` hook `pretty-format-json` now tries to reformat `ipynb` files, which causes our `pre-commit` hooks to fail.

Perhaps the right solution for us is to exclude `ipynb` from `pretty-format-json`. However, I wanted to mention that this change might trigger some ripple effects for users of `pre-commit`.

connortann:
I do think it makes sense for `ipynb` files to be included by default by hooks such as `check-json`, given that they are indeed JSON files. I don't know about `pretty-format-json`.

As far as I know, the `pretty-format-json` hook is compatible with Jupyter notebooks, but does reformat the files to a "pretty" indentation. Jupyter notebooks are [usually ](https://github.com/jupyter/nbformat/blob/4c265b76f7d445770f117334fc83874b9946f538/nbformat/v4/nbjson.py#L50C29-L50C29) saved with 1 space indentation, wherease `pretty-format-json` enforces 4 spaces by default.

I tried running the `pretty-format-json` hook with `args: ['--indent=1']` against a Jupyter notebook, and it passed without changes.

I suppose the question is whether one would _expect_ a jupyter notebook to be affected by `pretty-format-json`. Should it have a default of:

```yaml
types: [json]
exclude_types: [jupyter]
```

?

FelixSchwarz:
I treat jupyter files as "opaque" so I don't expect these to be formatted in any human readable way. We export jupyter files to python (`jupyter nbconvert`) so I can diff just the `.py` files. But that might be just my workflow.

Anyway in our case we had `jupyter nbconvert --clear-output --inplace` (which also part of our pre-commit config) fighting with `pretty-format-json` so that we could not get a state where both tools were happy with the result.

Personally I'd support an exclusion of `jupyter` for `pretty-format-json`, that seems like a good solution to me.

connortann:
Sounds sensible - I'll make a PR, given that this one caused you the issues in the first place :)

connortann:
@FelixSchwarz FYI that proposal was [rejected ](https://github.com/pre-commit/pre-commit-hooks/pull/937#issuecomment-1651871826) by the maintainers.

FelixSchwarz:
@connortann Thanks for trying :-)

PR #403, opened 2023-08-03
Support file.dockerfile files

mathbunnyru:
Fix: https://github.com/pre-commit/identify/issues/229

Some people create `<FILE>.dockerfile` (lowercase `.dockerfile`)  files, which are not currently parsed correctly.
It would be nice to support these files in `identify`, so `pre-commit` hooks work on these files as well.


asottile:
please don't send pull requests for declined features

PR #404, opened 2023-08-10
Add Fluent files extension (`.ftl`)

mondeja:
See [Fluent project](https://projectfluent.org/) and [an example of `.ftl` files usage](https://github.com/XAMPPRocky/fluent-templates#looking-up-fluent-resources).


asottile:
googling this extension does not land on fluent but instead freemarker templates -- as such it cannot be accepted

PR #413, opened 2023-09-14
Support identifying copy.bara.sky

sfc-gh-wzhao:
Add `copy.bara.sky` which is the default config file of [copybara](https://github.com/google/copybara). It is a starlark file like `.bzl`.

Not adding `.sky` extension as I did not see any other starlark file use `.sky` as extension.


asottile (approved):
it might make sense to add a starlark file type at some point -- but for now this seems fine

PR #418, opened 2023-10-03
Add extensions for Theos (Apple tweak framework) Logos preprocessor

jevinskie:
[Theos](https://theos.dev) is a framework that a majority of tweaks for jailbroken and non-jailbroken Apple devices use. It uses [Logos](https://github.com/theos/logos), a macro preprocessor written in Perl, to preprocess Objective-C/C++ files with `%token` macro extensions. The file extenions used are `.x`, `.xi`, `.xm`, and `.xmi` as documented [here](https://theos.dev/docs/logos-file-extensions). I identify them as `theos-logos` because Theos is the framework as a whole and Logos could be ambiguous or obscure.

I'd like to add these file extensions to `identify` because I'm working on a [`clang-format` wrapper](https://github.com/jevinskie/pre-commit-clang-format-theos) (or possibly adding support directly to Clang's LibFormat) to prevent `clang-format` from mangling the `%macro` tokens. I would have just used `types_or` with `c`, `c++`, `c`, `objective-c`, `objective-c++`, and `file` while specifying the additional Logos extensions with a `files` pattern but that seems to be an explicitly unsupported use case of `pre-commit` and I'd have to resort to `files` patterns for the original `clang-format` hook's supported files as well as the Logos files.


asottile:
these extensions are ambiguous so I cannot accept the patch. thanks anyway!

PR #424, opened 2023-11-02
Add 'typescript' alias extension for 'ts'

thiagowfx:
Precedent: JavaScript has both 'javascript' and 'js', this ensures consistency.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -241,7 +241,7 @@ EXTENSIONS = {
     'thrift': {'text', 'thrift'},
     'tiff': {'binary', 'image', 'tiff'},
     'toml': {'text', 'toml'},
-    'ts': {'text', 'ts'},
+    'ts': {'text', 'ts', 'typescript'},
     'tsv': {'text', 'tsv'},
     'tsx': {'text', 'tsx'},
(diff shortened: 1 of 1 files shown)


asottile:
there's no `js` tag so there's no precedence

thiagowfx:
Maybe my point was misunderstood. https://github.com/pre-commit/identify/blob/main/identify/extensions.py

What I meant was this:

Status quo:
- 'javascript' (no 'js' indeed as you say)
- 'ts' (no 'typescript')

With this PR, status quo:
- 'javascript'
- 'typescript'
- 'ts'

The rationale is: When using pre-commit with a kind filter, it has to be written like this: https://github.com/GoogleChromeLabs/chromium-bidi/blob/fc2d6847cc4c5dd10a46693055c7faa7827079f4/.pre-commit-config.yaml#L111

```
        types_or: [javascript, ts, python]
```

However, one would expect the following to be valid (but it's not!):

```
        types_or: [javascript, typescript, python]
```

...this PR intended to fix this inconsistency.

asottile:
search the issue tracker -- this inconsistency has been discussed before with resolution as to why it is not changing

PR #428, opened 2023-11-16
Add `empty` tag for zero-length files

bgilbert:
I have a local pre-commit hook using `pygrep --negate` to verify that Python files include a particular `__future__` import, and would like to exclude empty `__init__.py` files from the check.

For reference, `file` also has a special case for zero-length files:

```console
$ file __init__.py 
__init__.py: empty
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ If you have an actual file on disk, you can get the most information possible
 {'file', 'text', 'python', 'non-executable'}
 >>> identify.tags_from_path('/path/to/file-with-shebang')
 {'file', 'text', 'shell', 'bash', 'executable'}
+>>> identify.tags_from_path('/path/to/__init__.py')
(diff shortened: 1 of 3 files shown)


asottile:
this unacceptably regresses performance because it requires opening every file instead of identifying by name

asottile:
please in the future discuss new features before opening a PR and spending time on it -- include your rationale and use case when doing so as well

bgilbert:
Hmm, how will this degrade performance?  The test uses the information from the `stat` call that `tags_from_path()` already makes.

I think I gave the rationale in the PR description; anything more needed there?

asottile:
you'd probably be better to use something like `reorder-python-import`'s [--add-import](https://github.com/asottile/reorder-python-imports#adding--removing-imports) which does what you want

asottile:
you could also use lookahead / lookbehind in pygrep to check an empty file directly

PR #435, opened 2023-12-07
Add support for .cff (Citation File Format) extension as YAML

### LINKED ISSUE #434: Recognize Citation File Format (`.cff`) files as YAML

MattF-NSIDC:
Resolves #434

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -22,6 +22,7 @@ EXTENSIONS = {
     'c++m': {'text', 'c++'},
     'cc': {'text', 'c++'},
     'ccm': {'text', 'c++'},
+    'cff': {'text', 'yaml'},
     'cfg': {'text'},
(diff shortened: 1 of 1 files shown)


asottile:
searching this extension leads to ambiguity so unfortunately it cannot be accepted

MattF-NSIDC:
Understandable!

For future searchers looking to do this: for my project it was not straightforward / I didn't know how / have time to figure out how to get pre-commit to run prettier on my CITATION.cff file without also running it on a bunch of other stuff I don't want it to, like Markdown, or having to maintain a regex to exclude all the other possibilities identified as "text".

I ended up using the [cffconvert](https://github.com/citation-file-format/cffconvert) pre-commit hook which is designed specifically to validate this file. I decided to live without prettier :shrug:

PR #436, opened 2023-12-07
Detect ScalaFmt config as a scala file

danio:
Add ScalaFmt config so that pre-commit hooks will check changes to the config file and get latest

Currently when I run precommit on a scala repository using a hook that has:
`types: [scala]`
set in .pre-commit-hooks.yaml and I have local edits to my .scalafmt.conf file, precommit doesn't run the hook script as it doesn't find any scala files that have changed.

Identifying .scalafmt.conf as a scala file will fix this

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -331,6 +331,7 @@ NAMES = {
     '.pypirc': EXTENSIONS['ini'] | {'pypirc'},
     '.rstcheck.cfg': EXTENSIONS['ini'],
     '.salt-lint': EXTENSIONS['yaml'] | {'salt-lint'},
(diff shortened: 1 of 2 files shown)


danio:
> > When would a .conf file ever be anything other than a configuration file?
> > […](#)
> > On Thu, 7 Dec 2023 at 18:17, Anthony Sottile _**@**_._**> wrote: _**@**_.**_* commented on this pull request. ---------> 
> I think you're confused -- you've set `.conf` as `ini`

no need to be patronising!

reverted the change, but I think a config file type would be a useful concept, then ini would be a type of config and it would be clearer

PR #441, opened 2024-02-08
feat: Indentify `metal` as Metal Shading Language

Maetveis:
Metal Programming guide: Metal Tools

> ## File Extension for Metal Shading Language Source Files
>
> For Metal shading language source code file names, you must use the
> .metal file name extension to ensure that the development tools
> (Xcode and the GPU frame debugger) recognize the source files when
> debugging or profiling.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -151,6 +151,7 @@ EXTENSIONS = {
     'md': {'text', 'markdown'},
     'mdx': {'text', 'mdx'},
     'meson': {'text', 'meson'},
+    'metal': {'text', 'metal'},
     'mib': {'text', 'mib'},
     'mjs': {'text', 'javascript'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -151,6 +151,7 @@
     'md': {'text', 'markdown'},
     'mdx': {'text', 'mdx'},
     'meson': {'text', 'meson'},
+    'metal': {'text', 'metal-shading-language'},
```
asottile, line 154: it's more common now ot have the identify tags match the extension itself rather than something else

Maetveis:
> I can't fix your PR because you made it from a organization -- can you fix it for me?

I changed the file, should I rewrite the commit message too?

asottile:
right now you have two commits with identical messages

PR #442, opened 2024-02-10
Add xctestplan file type

KORraNpl:
For reference: https://github.com/pre-commit/identify/issues/432

While adding the new file type, I noticed some entries are not properly sorted, I've fixed it in a separate commit in case it's an unwanted change.


KORraNpl:
Removed fixed sorting, as requested.

asottile (approved):

PR #443, opened 2024-02-15
Add typst extension

pavelzw:
https://github.com/typst/typst uses `.typ` as a file extension


asottile:
according to https://github.com/github-linguist/linguist/issues/6358 the `.typ` extensions is potentially ambiguous so it unfortunately cannot be accepted here

PR #453, opened 2024-04-10
add-elixir-file-extensions

arcanemachine:
Adds the `eex` and `heex` Elixir file extensions


identify/extensions.py
```diff
@@ -90,6 +91,7 @@
     'h': {'text', 'header', 'c', 'c++'},
     'hbs': {'text', 'handlebars'},
     'hcl': {'text', 'hcl'},
+    'heex': {'text', 'elixir'},
```
asottile, line 94: these appear to have a different syntax than base elixir files -- they should probably be handled separately

PR #455, opened 2024-04-15
Identify Cargo.toml and Cargo.lock as cargo and cargo-lock

jalaziz:
Introduce dedicated types for Cargo.toml and Cargo.lock files. These are similar to go.mod and go.sum files.


jalaziz:
I separated out `Cargo.toml` and `Cargo.lock` as separate types to match `go.mod` and `go.sum`, but maybe they should just be grouped under a single `cargo` type?

asottile (approved):

PR #457, opened 2024-04-25
Add ejson extension

wircho:
`ejson` is a pretty popular "encrypted json" format and open source tool built by Shopify. `ejson` files are just JSON files where terminal string nodes are encrypted.

[More info](https://github.com/Shopify/ejson)

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -56,6 +56,7 @@ EXTENSIONS = {
     'ear': {'binary', 'zip', 'jar'},
     'edn': {'text', 'clojure', 'edn'},
     'ejs': {'text', 'ejs'},
+    'ejson': {'text', 'json', 'ejson'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -56,6 +56,7 @@
     'ear': {'binary', 'zip', 'jar'},
     'edn': {'text', 'clojure', 'edn'},
     'ejs': {'text', 'ejs'},
+    'ejson': {'text', 'json'},
```
asottile, line 59: I would also add `ejson` so this can be targetted specifically in addition to `json`

wircho: Done!

wircho: @asottile Do I need to do something else or will you take care of merging this? 🙏

asottile (approved):

PR #468, opened 2024-06-27
Fix IndexError when shebang is just '#!/usr/bin/env'

harrymander:


asottile:
is there an actual case where this came up? feels a bit like garbage-in-garbage-out

harrymander:
I did come across it randomly due to a typo in a file.  I agree that it is a pretty unlikely scenario, though it seems to be the only case where the library raises an exception due to the contents of the file. I suppose it's not too hard for client code to just check for this edge case if it may be a problem

identify/identify.py
```diff
@@ -192,7 +192,7 @@ def parse_shebang(bytesio: IO[bytes]) -> tuple[str, ...]:
 
     cmd = tuple(_shebang_split(first_line.strip()))
     if cmd and cmd[0] == '/usr/bin/env':
-        if cmd[1] == '-S':
+        if len(cmd) > 1 and cmd[1] == '-S':
             cmd = cmd[2:]
         else:
             cmd = cmd[1:]
```
asottile, line 198: I think this can probably be written a little simpler:

```python
if cmd[:2] == ('/usr/bin/env', '-S'):
    return cmd[2:]
elif cmd[:1] == ('/usr/bin/env',):
    return cmd[1:]
elif cmd == ('nix-shell',):
    return _parse_nix_shebang(bytesio, cmd)
else:
    return cmd
```

asottile (approved):

PR #470, opened 2024-07-08
Add CAPL .can extension

lalten:
Recognize .can files to be [CAPL](https://www.vector.com/us/en/know-how/capl) script. It looks just like C and can be formatted by clang-format.


asottile:
searching this extension leads to many different things so unfortunately this fails the unambiguity requirement

PR #473, opened 2024-08-04
Use Assignment Expression (Walrus) In Conditional

havocesp:
Suggest accepted, so merging branches.


asottile:
please don't spam this ai garbage anywhere again

PR #474, opened 2024-08-04
Interpreters updated and a minor code change

havocesp:
I have added most of python 2 and 3 versions and some other interpreters like xonsh, sed, or xdotool (found sometimes as shebangs).


asottile:
please only change one thing in a patch, don't change unrelated style, and generally the things you're changing don't improve things

PR #478, opened 2024-08-29
Add just (https://just.systems/)

starthal:
Just is a command runner.

https://just.systems

Include the conventional justfile and module names

Include the interpreter heuristic: https://just.systems/man/en/chapter_59.html?highlight=scrips#just-scripts


asottile:
please search for duplicates

starthal:
@asottile Apologies, I did search and found https://github.com/pre-commit/identify/pull/300 but I'm not clear on why it was closed. It seemed like it might have been due to staleness.

PR #479, opened 2024-09-30
Support fmf

imphil:
fmf ("Flexible Metadata Format") is a YAML-based file format, mainly used to
describe tests with the TMT framework. TMT is used by RHEL-based distributions
for integration testing of packages.

See also https://fmf.readthedocs.io/en/stable/concept.html#choices.


asottile:
a search for this file extension is ambiguous so unfortunately it cannot be accepted

PR #490, opened 2024-12-02
feat: identify typ files as typst

AustinZhang1024:
Add support for typst files.


asottile:
please search for duplicates

PR #491, opened 2024-12-10
Add .robot and .resource to extensions.py

djalan:
Hello

Please kindly add `.robot` and `.resource` extensions for Robot Framework. They are text files containing *pretty code* formulated in English that actually run Python code.

They look a lot like `.feature` files that use the Gherkin syntax.

Here is a Robot file:

```robot
*** Settings ***
Documentation     A test suite for valid login.
...
...               Keywords are imported from the resource file
Resource          keywords.resource
Default Tags      positive

*** Test Cases ***
Login User with Password
    Connect to Server
    Login User            ironman    1234567890
    Verify Valid Login    Tony Stark
    [Teardown]    Close Server Connection
```


asottile:
extensions need to be unambiguous of which neither of these appear to be from a quick Internet search

djalan:
Thanks for your quick reply.

Robocorp raised 32M$ in the last 5 years and Robot Framework is the biggest Python automation framework.

I just searched `.robot file extension` on Google and DuckDuckGo and the no. 1 result is what we are looking for.

Moreoever, `.rb` is already taken by the Ruby programming language. `.robot` is a logicial choice.

But, I have to agree with you that `.resource` was a horrendous decision. It should have been `.robotr` or something like that. I searched Google for `filetype:resource site:github.com` and there are many conflicts.

I could amend the PR to only add the `.robot` extension and get rid of `.resource`.



@pekkaklarck would you like to comment?

pekkaklarck:
Robot Framework is fairly popular and I don't know about any other project using the `.robot` extension, so including it in a list of known extensions would sound fair for me. Anyway, if maintainers see otherwise, I don't have much interest to argue. Their project, their rules.

This isn't the correct place to discuss is the `.resource` extension good or not. That said, I doubt changing it now would be worth the effort, but we could extend it so that something like `example.robot.resource` would work so that `example` would be considered the base name. If you think that would make sense, please submit an issue to Robot's tracker.

PR #494, opened 2024-12-16
Add Tiltfile

dbast:
Add Tiltfile ~~+ Starlark~~ support. Tiltfiles are written in Starlark, which is a language intended for use as a configuration language. It was originally designed for the [Bazel](https://bazel.build/) build system. Starlark is a dialect of **Python**.

~~background: Lots of Python targeting pre-commit hooks (like ruff) can also be applied to Starlark files. This PR enables to apply those hooks via specifying `types_or: [starlark]` or  `types_or: [tiltfile]` or `types_or: [bazel]` depending on what is needed. Not every Python pre-commit hook applies to every Starlark file, as there are different variants with some [language features](https://pkg.go.dev/go.starlark.net/resolve?utm_source=godoc#pkg-variables) enabled or disabled.~~

refs:
* https://docs.tilt.dev/api
* https://github.com/bazelbuild/starlark

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -393,6 +393,7 @@ NAMES = {
     'setup.cfg': EXTENSIONS['ini'],
     'sys.config': EXTENSIONS['erl'],
     'sys.config.src': EXTENSIONS['erl'],
+    'Tiltfile': {'text', 'tiltfile'},
     'Vagrantfile': EXTENSIONS['rb'],
     'WORKSPACE': EXTENSIONS['bzl'],
     'wscript': EXTENSIONS['py'],
--- a/tests/identify_test.py
+++ b/tests/identify_test.py
@@ -172,6 +172,9 @@ def test_tags_from_path_plist_text(tmpdir):
         ('meson.build', {'text', 'meson'}),
         ('meson_options.txt', {'text', 'plain-text', 'meson'}),
         ('Vagrantfile', {'text', 'ruby'}),
+        ('Tiltfile', {'text', 'tiltfile'}),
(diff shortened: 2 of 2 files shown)


asottile:
so there is an odd precedent set in #413 that treats all starlark as "bazel" -- I don't think that's quite correct but then this should get resolved somehow so this PR isn't inconsistent

we can't _remove_ bazel from that extension so maybe the best is to just add it?

dbast:
@asottile rebased the PR and also no explicitly setting `'copy.bara.sky': {'text', 'bazel', 'starlark'}` (it implicitly already was like that before as it referenced `blz` which also is `{'text', 'bazel', 'starlark'}`). So it remains backward compatible, but downstream users can now also use `starlark` (instead of `bazel`), if they apply to all starlark files.

dbast:
Looking at the `files` regex in https://github.com/keith/pre-commit-buildifier/blob/main/.pre-commit-hooks.yaml tell that there are maybe more files/extension that can be added to be identified as Starlark to simplify such kind of hooks or to use existing python hooks to also apply to starlark files. This PR focuses on adding Tiltfile detection and marking the existing definitions that are starlark to be also be detected as such. Further detections and improvements can be done via further PRs.

@asottile anything else to do here as part of this PR? thanks!

asottile:
I'm hesitant to add `starlark` at this time -- at least not in this PR which is currently doing multiple things.  I'd rather have one way to do things tbh

dbast:
ok. stripped the PR down to only add Tiltfile support. Starlark can be added in the future via further PRs if desired to existing and new files/extension.

asottile (approved):

PR #496, opened 2024-12-17
feat: Add support for /bin/env

### LINKED ISSUE #495: Does not identify files using /bin/env in shebang line
identify fails to properly identify files which use `#!/bin/env` instead of `#!/usr/bin/env`:

For example:

* `#!/usr/bin/env python`:
```
identify-cli usr-bin-env-python-test
["executable", "file", "python", "text"]
```

* `#!/bin/env python`:
```
identify-cli bin-env-python-test
["executable", "file", "text"]
```

psa:
Some places use /bin/env instead of /usr/bin/env for interpreted scripts. Support both.

Closes #495


asottile:
this was declined.  some places are wrong and should follow the established convention

PR #501, opened 2025-01-14
feat: add `templ` to extensions.py

dinoshauer:
Adds `.templ` as an extension as per https://templ.guide/core-concepts/template-generation

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -250,6 +250,7 @@ EXTENSIONS = {
     'swiftdeps': {'text', 'swiftdeps'},
     'tac': {'text', 'twisted', 'python'},
     'tar': {'binary', 'tar'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -249,6 +249,7 @@
     'swiftdeps': {'text', 'swiftdeps'},
     'tac': {'text', 'twisted', 'python'},
     'tar': {'binary', 'tar'},
+    'templ': {'text', 'go', 'templ'},
```
asottile, line 252: I'm not sure the `go` tag is appropriate here -- the source doesn't look like go syntax? would tools that expect go files be able to accept templ files?

dinoshauer: Ah no, you're right. I misunderstood the format :) I'll remove `go`

PR #503, opened 2025-01-30
add uv to interpreters list

mheiges:
add [uv](https://github.com/astral-sh/uv) to the interpreters list


identify/interpreters.py
```diff
@@ -18,6 +18,7 @@
     'python': {'python'},
     'python2': {'python', 'python2'},
     'python3': {'python', 'python3'},
+    'uv': {'python', 'python2', 'python3'},
```
asottile, line 21: the versioned tags never really caught on -- I'm going to remove them at some point.  also uv doens't run python 2 code as far as I can tell

asottile (approved):

asottile:
#512

PR #508, opened 2025-02-12
Add vyper

romanagureev:
[Vyper](https://vyperlang.org) – pythonic language for smart contracts.


romanagureev:
@asottile Where does ambiguity come from? Could not google same extensions.

PR #514, opened 2025-03-15
feat: jsonc support

stevenh:
Add support for identifying jsonc files by extension.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 EXTENSIONS = {
     'adoc': {'text', 'asciidoc'},
     'ai': {'binary', 'adobe-illustrator'},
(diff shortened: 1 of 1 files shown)


asottile:
afaik this isn't a standardized format or extension (does it include commas? does it include single quotes?) so we can't accept an extension for it

stevenh:
JSONC is a superset of JSON which supports comments, it's commonly used by tools like Visual Studio Code.

You can some more details here:
* https://github.com/JSONC-Spec/JSONC-Spec
* https://github.com/muhammadmuzzammil1998/jsonc

It's also supported by prettier, and I noticed the lack of support while doing a PR for [schemastore](https://github.com/SchemaStore/schemastore) you can see the [comment here](https://github.com/SchemaStore/schemastore/blob/master/.pre-commit-config.yaml#L7) which currently prevents it from using prettier to check jsonc files as far as I can tell.

Hope this helps, do you need any more information?

asottile:
the usage of the extension predates that "standard"

stevenh:
Is there a way to implement it such that we can get it added?

asottile:
you don't _need_ identify support to use things in pre-commit -- you can always use `files`

stevenh:
Thanks for the info, I'll take a look.

PR #515, opened 2025-03-25
Add matlab extension

da-phil:
I'm using a code-base with matlab *.m files together with pre-commit, identifying matlab files would be quite helpful ;)

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -154,7 +154,7 @@ EXTENSIONS = {
     'lpr': {'text', 'lazarus', 'pascal'},
     'lr': {'text', 'lektor'},
     'lua': {'text', 'lua'},
-    'm': {'text', 'objective-c'},
+    'm': {'text', 'objective-c', 'matlab'},
     'm4': {'text', 'm4'},
     'make': {'text', 'makefile'},
     'manifest': {'text', 'manifest'},


da-phil:
> extensions must be unambiguous

I don't understand. The extension dict is full of multiple associations, some extensions even have 4 associations. I thought this was the whole point of this dict, to create a mapping from a file extension to a programming language or tool. And because extensions can be used by different tools (e.g. .sh can be used by all shells) multiple mappings for the same extensions are to be expected.

asottile:
`.m` is not unambiguously matlab -- that extension is used by many other things and therefore we cannot accept it into the extensions list

there may be things in the list that are ambiguous -- unfortunately mistakes were made in the last decade of this existing and we're trying to not do that moving forward

da-phil:
In this example this would be only a problem if pre-commit would be used in a code-base were both objective-c and matlab files are present. A  `files` regex could be used to disambiguate by path (assuming that matlab and objective-c files are not mixed within the same folders), but then `types` can be still used for the respective hooks if one wouldn't want to have the file extension part of the `files` regex.
I mean, it's not a big deal to just deal with it in the `files` regex, but I really like the readability of something like `types: [matlab]` or `types: [objective-c]`  and keeping the `files` regex as simple as possible.

I don't think its a fatal design flaw if `types: [matlab]` and `types: [objective-c]` lead to the same filtering behavior for a hook. That's just a mirror of the reality of having the same extension for different interpreters and tools.

asottile:
look.  I don't have time to debate you.  I am telling you how it is.

PR #519, opened 2025-05-21
Identify common MSBuild files

gtbuchanan:
https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild
> MSBuild does not change its behavior based on the file extension, but the convention is to use extensions ending in proj such as .csproj, .vcxproj, or .vbproj, for the main MSBuild input file, called the "project file."

> The Visual Studio build system stores project-specific logic in the project file itself, and uses imported MSBuild XML files with extensions like .props and .targets to define the standard build logic. The .props files define MSBuild properties, and .targets files define MSBuild targets.

I didn't include _all_ known types (like the other `**proj` listed [here](https://stackoverflow.com/a/2012217/1409101)), but the more common ones should be sufficient until someone needs the others.


asottile:
> I was afraid of that. I assume you're referring to `proj` and `props`?

tbh I stopped looking once I saw proj so there might be more

PR #522, opened 2025-05-23
Identify LLVM TableGen files

TomMurray:
https://llvm.org/docs/TableGen/ProgRef.html#source-files
> The standard file extension for TableGen files is .td.


TomMurray:
> please search for duplicates in the future !

Ah apologies my bad!

PR #527, opened 2025-06-30
Add some common file extensions.

limburgher:
The image, audio, and video types are fairly straightforward. A few are audio and video, I hope I handled that reasonably.


asottile:
please make sure these are unambiguous

PR #535, opened 2025-08-22
fix: add robot filetype and sort extensions

mike325:


asottile:
do one thing per pr please

PR #539, opened 2025-09-21
Add names of rust linter and formatter config files

BaumiCoder:
In Rust there is a default formatter, called `rustfmt`, and a default linter, called `clippy`. This Pull Request adds the names (with and without leading dot) of their config files (see commit message for links to the respective documentations).
Furthermore, the single file name for `cargo-deny`, a linter focused on cargo / Rust dependencies.

(Useful, when running them with `pre-commit` and only want to run, if relevant files have changed, i.e., rust files, `Cargo.toml` or the config file in the repository. Example `pre-commit` config for a Rust project [here](https://codeberg.org/BaumiCoder/ecformat/src/branch/feature/2-linters/.pre-commit-config.yaml))

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -374,6 +374,7 @@ NAMES = {
     'BUILD': EXTENSIONS['bzl'],
     'Cargo.toml': EXTENSIONS['toml'] | {'cargo'},
     'Cargo.lock': EXTENSIONS['toml'] | {'cargo-lock'},
+    'deny.toml': EXTENSIONS['toml'] | {'cargo-deny'},
     'clippy.toml': EXTENSIONS['toml'] | {'clippy'},
     'CMakeLists.txt': EXTENSIONS['cmake'],
     'CHANGELOG': EXTENSIONS['txt'],


asottile:
I don't see how this can be useful?  why wouldn't this just use `files` to match these?  they're already matched by toml without special conventions

BaumiCoder:
With that changes, for example I would use this for `rustfmt`:
```yml
types_or: [cargo, rust, rustfmt]
```
I am not sure, what your alternative approach with the same filtering would be. For example this would not match anymore the `*.rs` files, because there is an "and" between `files` and `types_or`, isn't it?
```yml
files: "\\.?rustfmt\\.toml"
types_or: [cargo, rust, toml]
```
So everything, needs to be defined with `files` in this way,
```yml
files: "(\\.?rustfmt\\.toml|Cargo.toml|.+\\.rs)"
```
which is not that readable.

Only using `types_or` would also include other TOML files, which are not relevant for `rustfmt`
```yml
types_or: [cargo, rust, toml]
```

asottile:
that doesn't really make sense, you would just use `types: [rust]` such that pre-commit passes just the rust files to the hook

BaumiCoder:
I call it on the whole repository:
```yaml
      - id: rustfmt
        name: rustfmt
        description: The Rust default formatter (see .rustfmt.toml for config)
        # Run via cargo to use edition from Cargo.toml
        entry: cargo fmt --check --verbose
        language: rust
        types_or: [cargo, rust]
        pass_filenames: false
```
Running `rustfmt` via cargo allows getting the `edition` property from the `Cargo.toml`. Furthermore, I want to run it also when it config changes (i.e., the `rustfmt.toml`) to avoid that a config change makes the formatting invalid in respect to the new config.

Similar is for the two linter, which are also called on the whole repository and config changes are relevant.
```yaml
      - id: cargo-deny
        name: cargo-deny
        description: A Linter which checks Rust dependencies (e.g., regarding licenses)
        entry: cargo deny check
        language: rust
        types_or: [cargo, cargo-lock]
        pass_filenames: false
        additional_dependencies: [cli:cargo-deny:0.18.*]
      - id: clippy
        name: clippy
        description: The Rust default linter
        entry: cargo clippy
        language: rust
        types_or: [cargo, rust]
        pass_filenames: false
        stages: [pre-push]
```

asottile:
> I call it on the whole repository

yeah I'd say you're holding it wrong then.  part of the point of pre-commit is to only run on the things that change

asottile:
no I wouldn't bother with a "run everything" hook.  just do that manually and you should be running everything in ci already anyway.

BaumiCoder:
In my CI (at Codeberg) I plan to use `pre-commit run --all-files -v --hook-stage manual`. This avoids a separate config for linters (and their versions) as pre-commit handles the installation of all necessary linters for the CI.

PR #541, opened 2025-09-23
Add .piskel extension

wjt:
Piskel is a browser-based sprite editor. Its file format is JSON-based (though the JSON contains base64-encoded PNG data).

Recognise the .piskel extension and identify it analogously to .svg: it is text, image, json and piskel.

https://github.com/piskelapp/piskel
http://justsolve.archiveteam.org/wiki/Piskel_canvas

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -194,6 +194,7 @@ EXTENSIONS = {
     'php4': {'text', 'php'},
     'php5': {'text', 'php'},
     'phtml': {'text', 'php'},
+    'piskel': {'text', 'piskel', 'json'},
     'pl': {'text', 'perl'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -194,6 +194,7 @@
     'php4': {'text', 'php'},
     'php5': {'text', 'php'},
     'phtml': {'text', 'php'},
+    'piskel': {'text', 'image', 'piskel', 'json'},
```
asottile, line 197: the `image` tag was an idea that didn't really pan out so I'm just going to remove it here

wjt: That's a shame, because in this case the image tag is just what we want!

PR #549, opened 2025-11-13
Provide support for D and VALA language files

### LINKED ISSUE #258: Add .d and .vala as filetypes
Both D and vala are languagse and have `.vala` and `.d` filetypes respectively. They are text files.

ngie-eign:
This change adds file extension detection support for .vala files, which is
used by the VALA programming language.

Closes:	#258


asottile:
unfortunately I don't think `.d` is unambiguous

PR #555, opened 2025-12-30
Add .ipxe extension

Roxedus:
Also managed to sort some of the extensions starting with I, as this repo lacks a contributing guide, I left this sorting be.


asottile (changes requested):
please only make one change per PR

asottile (approved):

PR #556, opened 2026-01-04
Add HIP '.hip' file extension

lijinpei-amd:


asottile:
unfortunately this extension is ambiguous and so it cannot be accepted

PR #559, opened 2026-01-28
Add support for Pkl

pjjw:
Pkl is a configuration language from Apple: https://pkl-lang.org/index.html


asottile:
search next time please

PR #560, opened 2026-01-30
Add 'escript' file type and interpreter for Erlang

sebastiw:
Hope I understand it correctly.
File types with `my_script.escript` should be added as Erlang, as well as when the file type is missing it can be matched by the shebang `#!/usr/bin/env escript`  

https://www.erlang.org/doc/apps/erts/escript_cmd

Thank you

### DIFF
--- a/identify/interpreters.py
+++ b/identify/interpreters.py
@@ -8,6 +8,7 @@ INTERPRETERS = {
     'csh': {'shell', 'csh'},
     'dash': {'shell', 'dash'},
     'expect': {'expect'},
+    'escript': {'erlang'},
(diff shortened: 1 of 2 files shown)


identify/interpreters.py
```diff
@@ -8,6 +8,7 @@
     'csh': {'shell', 'csh'},
     'dash': {'shell', 'dash'},
     'expect': {'expect'},
+    'escript': {'shell', 'erlang'},
```
asottile: `shell` is usually reserved for posixlike shells (`sh` / `bash` / etc.) -- you probably just want `erlang` here

PR #563, opened 2026-02-04
Add support for sconstruct and sconscript extensions

seanbudd:
Scons scripting support: https://scons-cookbook.readthedocs.io/en/latest/#sharing-variables-between-sconscript-files

These are files named sconstruct or sconscript. 
It might be worth including them with python, but that might cause some tools to break by default, as it's not normal python code.


identify/extensions.py
```diff
@@ -411,6 +411,8 @@
     'Rakefile': EXTENSIONS['rb'],
     'rebar.config': EXTENSIONS['erl'],
     'setup.cfg': EXTENSIONS['ini'],
+    'sconstruct': {'text', 'scons'},
+    'sconscript': {'text', 'scons'},
```
asottile: doesn't scons have a funny capitalization convention for these files?

PR #571, opened 2026-02-10
Add support for *.entitlements

petamas:
Format documentation: https://developer.apple.com/documentation/bundleresources/entitlements

Relevant quote:
> Xcode records capabilities that you add in a property list file with the .entitlements extension.


identify/extensions.py
```diff
@@ -324,6 +324,7 @@
     'zsh': {'text', 'shell', 'zsh'},
 }
 EXTENSIONS_NEED_BINARY_CHECK = {
+    'entitlements': {'plist', 'entitlements'},
```
asottile: can this actually be a binary file?  all the examples I see online are either empty or xml

PR #572, opened 2026-02-10
Add support for *.hlsl and *.hlsli

petamas:
Documentation: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl

I did not find any direct source in the above docs saying that `.hlsl` is the standard extension, but Google's not giving me anything else, so it seem unambigiouos enough.

I've found this on the Stack Overflow answer on the `.hlsl` / `.hlsli` distinction though, and it matches existing practice I've seen in the wild: https://stackoverflow.com/questions/46809070/what-is-the-difference-between-hlsl-and-hlsli

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -103,6 +103,8 @@ EXTENSIONS = {
     'h': {'text', 'header', 'c', 'c++'},
     'hbs': {'text', 'handlebars'},
     'hcl': {'text', 'hcl'},
+    'hlsl': {'text', 'hlsl'},
+    'hlsli': {'text', 'hlsl'},
     'hh': {'text', 'header', 'c++'},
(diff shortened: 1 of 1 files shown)


identify/extensions.py
```diff
@@ -103,6 +103,8 @@
     'h': {'text', 'header', 'c', 'c++'},
     'hbs': {'text', 'handlebars'},
     'hcl': {'text', 'hcl'},
+    'hlsl': {'text', 'hlsl'},
+    'hlsli': {'text', 'header', 'hlsl'},
```
asottile: generally `header` is meant for c / c++ headers and really isn't that useful of a tag so I'm not adding it to new things

PR #593, opened 2026-05-11
Add support for 'luau' file extension

NinjaMandalorian:
Adds support for the `.luau` file extension, a filetype used most commonly around the Roblox ecosystem, as well as some standalone runtimes.

Hooks created for pre-commit (like [StyLua's](https://github.com/JohnnyMorganz/StyLua/blob/main/.pre-commit-hooks.yaml) from 5 years ago) are aimed at the "lua" type. Adding this allows for backwards compatibility.

### DIFF
--- a/identify/extensions.py
+++ b/identify/extensions.py
@@ -163,6 +163,7 @@ EXTENSIONS = {
     'lpr': {'text', 'lazarus', 'pascal'},
     'lr': {'text', 'lektor'},
     'lua': {'text', 'lua'},
+    'luau': {'text', 'lua', 'luau'},
     'm': {'text', 'objective-c'},
     'm4': {'text', 'm4'},
     'magik': {'text', 'magik'},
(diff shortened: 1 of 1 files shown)


asottile:
hmm it seems to me that `luau` is a superset of `lua` so I'm not sure it makes sense to tag it as `lua` (since a `luau` file may use features not expected by a tool targetting `lua`)

NinjaMandalorian:
> hmm it seems to me that `luau` is a superset of `lua` so I'm not sure it makes sense to tag it as `lua` (since a `luau` file may use features not expected by a tool targetting `lua`)

From my understanding, the deviation is more around adding typing, as luau is fully compatible with Lua 5.1, and does inherit features from future versions, [albeit not fully](https://luau.org/compatibility/#differences-from-lua).

I'll edit the PR either way to remove lua due to these differences.

PR #598, opened 2026-07-01
Run tests on Python 3.14 and 3.15

edgarrmondragon:
Noticed these were missing, hope it's ok


asottile:
don't touch people's infrastructure unless they ask for it

PR #599, opened 2026-07-06
Add TableGen Support

Naville:
This is LLVM's [TableGen](https://llvm.org/docs/TableGen/) DSL

TableGens seems to be supported by clang-format. So in theory once this is merged [pre-commit/mirrors-clang-format](https://github.com/pre-commit/mirrors-clang-format/blob/main/.pre-commit-hooks.yaml) should also be updated


Naville:
@asottile is there any reason why this is closed?

asottile:
yes, search

Naville:
ahh my bad

PR #601, opened 2026-07-15
Provide support for VALA language files

### LINKED ISSUE #258: Add .d and .vala as filetypes

ngie-eign:
This change adds file extension detection support for .vala files, which is used by the VALA programming language.

Closes:	#258

_(This is a resubmission of #549)_


pocc:
This relates to an active issue that https://github.com/pocc/pre-commit-hooks/issues/67 depends on.

ngie-eign:
@asottile : ping?

